Issue #4, opened 2022-02-05
Adding a Stream ID


FragLegs:
In previous discussion, it was decided that we needed a way for there to be potentially more than one stream between a transmitter and receiver. Some use cases for this were the needs of GDPR to keep data localized, having different polling frequencies for different types of events, etc. What follows is a high-level proposal for how we could add stream IDs without significantly changing the structure of the SSE framework. I would like to have some discussion here so that the high-level details can be settled before I create a pull request for the change.

## Support for default and/or named streams
First of all, because stream IDs dictate something about how the data model works, transmitters should be able to decide whether they support streams with IDs or not. Currently, there is a single default stream per receiver. I propose adding a new optional property to the [transmitter configuration response](https://github.com/openid/sse/blob/main/openid-sse-framework-1_0.txt#L680-L699) from `/.well-known/sse-configuration`:
```
stream_types: [
  "https://schemas.openid.net/secevent/stream_type/default",
  "https://schemas.openid.net/secevent/stream_type/named"
]
```
This allows the transmitter to support both default and named streams. Many, perhaps most, receivers will only care about having a single stream, so continuing to support a default stream is user friendly.

The default stream does not need to be created. It exists as soon as there is a transmitter/receiver relations if the transmitter supports default streams. Any named streams must be explicitly created.

### Changes to the stream configuration object
The most significant changes to the spec are centered around the [configuration_endpoint](https://github.com/openid/sse/blob/main/openid-sse-framework-1_0.txt#L1111), in the [GET](https://github.com/openid/sse/blob/main/openid-sse-framework-1_0.txt#L1188), [POST](https://github.com/openid/sse/blob/main/openid-sse-framework-1_0.txt#L1307) and [DELETE](https://github.com/openid/sse/blob/main/openid-sse-framework-1_0.txt#L1482) verbs, as well as in a new PATCH verb (detailed below).

With the exception of DELETE, all of these verbs either take a Stream Configuration object as an argument or return a Stream Configuration object. In all cases, we would update the Stream Configuration object to include a new optional parameter, `stream_id`:
- stream_id (if missing, assume default stream)
- iss (read only)
- aud (read only)
- events_supported (read only)
- events_delivered (read only)
- min_verification_interval (read only)
- events_requested (read/write)
- delivery (read/write)
- format (read/write)

In some ways, you could think of a default stream as a stream whose `stream_id` is `null`.

### Create a Stream - configuration_endpoint [POST]
Creates a new stream

#### Arguments
[...]

PR #9, opened 2022-05-04
Add Stream ID

FragLegs:
This PR does a number of things:

- Most notably, it implements the changed described in issue #4 - Adding a Stream ID
- It reorders the sections so that information about creating/updating/deleting the stream comes before information about setting/getting the stream status
- It adds Shayne Miel as a coauthor
- It adds some info to the README to help others who might work on the spec


FragLegs:
However, there is a slight wrinkle with the stream ID that I didn't foresee. How should we indicate to transmitters and receivers that they need a way to identify which stream is being POLLed or PUSHed? 

For PUSH it is easy - we can leave it up to the Receiver to set either a URL param in the endpoint or use a specific endpoint that dictates the stream ID. It is a little complicated though since the Receiver does not know the stream ID when creating the stream, which would force them to provide the "delivery" info during a second request. How much of this should we call out in the spec?

For POLL it is up to the Transmitter to do the right thing and add a URL param or a specific endpoint for each stream. Do we need to call that out in the spec?

A third option would be to extend the POLL spec to include a stream_id parameter, but that seems like a big lift.

adeinega:
As an (almost) completely random comment, it would be great to keep only *.md documents in the repo, for example, openid-sse-framework-1_0.md, and then

1. use https://github.com/cabo/kramdown-rfc to convert *.md to *.xml
2. convert *.xml using `xml2rfc` to txt and HTML files when you they need to be published in other (external) places

It'll allow keeping the number of changes very low and easily review them. Moreover, Markdown files are convenient to edit even though the Web UI of GitHub.

timcappalli:
> As an (almost) completely random comment, it would be great to keep only *.md documents in the repo, for example, openid-sse-framework-1_0.md, and then
> 
> 1. use https://github.com/cabo/kramdown-rfc to convert *.md to *.xml
> 2. convert *.xml using `xml2rfc` to txt and HTML files when you they need to be published in other (external) places
> 
> It'll allow keeping the number of changes very low and easily review them. Moreover, Markdown files are convenient to edit even though the Web UI of GitHub.

that's the plan. Just haven't had a change to convert the others to MD and then set up GH actions

FragLegs:
> However, there is a slight wrinkle with the stream ID that I didn't foresee. How should we indicate to transmitters and receivers that they need a way to identify which stream is being POLLed or PUSHed?
> 
> For PUSH it is easy - we can leave it up to the Receiver to set either a URL param in the endpoint or use a specific endpoint that dictates the stream ID. It is a little complicated though since the Receiver does not know the stream ID when creating the stream, which would force them to provide the "delivery" info during a second request. How much of this should we call out in the spec?
> 
> For POLL it is up to the Transmitter to do the right thing and add a URL param or a specific endpoint for each stream. Do we need to call that out in the spec?
> 
> A third option would be to extend the POLL spec to include a stream_id parameter, but that seems like a big lift.

I added notes in the PUSH/POLL section to indicate the uniqueness needs of the URLs.

tulshi:
In the Transmitter Configuration Metadata, should we modify the "delivery
methods supported" to include the POLL URL if the method is POLL?

FragLegs:
That’s a good idea, but can we make it a separate issue/PR? I think there’s
enough to discuss there to warrant a separate issue.

mCloudCorp-sudo (approved):

openid-sse-framework-1_0.txt
```diff
@@ -557,39 +552,44 @@ Tulshibagwale, et al.   Expires December 10, 2021               [Page 9]
...
-   since the Issuer contains no path component:
+   https://tr.example.com to obtain its Configuration information, since
+   the Issuer contains no path component:
 
    GET /.well-known/sse-configuration HTTP/1.1
    Host: tr.example.com
 
     Figure 9: Example: Transmitter Configuration Request (without path)
 
-   If the Issuer value contains a path component, any terminating "/"
```
tulshi, line 587: I've noticed that the quotes are gone from many places in the spec. Is that intentional or an artifact of the tools you are using?

openid-sse-framework-1_0.txt
```diff
@@ -557,39 +552,44 @@ Tulshibagwale, et al.   Expires December 10, 2021               [Page 9]
...
                           openid-sse-framework                 June 2021
 
 
    Transmitters supporting Discovery MUST make a JSON document available
-   at the path formed by inserting the string "/.well-known/sse-
-   configuration" into the Issuer between the host component and the
-   path component, if any.  The syntax and semantics of ".well-known"
-   are defined in [RFC5785].  "sse-configuration" MUST point to a JSON
+   at the path formed by inserting the string /.well-known/sse-
+   configuration into the Issuer between the host component and the path
```
tulshi, line 567: I think the quotes are important here, otherwise the spec cannot be read unambiguously

openid-sse-framework-1_0.txt
```diff
@@ -707,29 +681,22 @@ Tulshibagwale, et al.   Expires December 10, 2021              [Page 12]
...
    Tokens issued from this Transmitter.
 
 7.  Management API for SET Event Streams
 
    This section defines an HTTP API to be implemented by Event
-   Transmitters and that can be used by Event Receivers to query and
-   update the Event Stream configuration and status, to add and remove
-   subjects and to trigger verification.
+   Transmitters which can be used by Event Receivers to create and
+   delete one or more Event Streams, query and update the configuration
```
tulshi, line 693: Since we are talking about multiple event streams now, we should probably write a new section that defines what an event stream is.

openid-sse-framework-1_0.txt
```diff
@@ -1448,47 +1397,211 @@ Tulshibagwale, et al.   Expires December 10, 2021              [Page 25]
...
+   The following is a non-normative example request to delete an Event
+   Stream:
+
+  DELETE /sse/stream?stream_id=f67e39a0a4d34d56b3aa1bc4cff0069f HTTP/1.1
+  Host: transmitter.example.com
+  Authorization: Bearer eyJ0b2tlbiI6ImV4YW1wbGUifQo=
+
+                Figure 22: Example: Delete Stream Request
+
+   Errors are signaled with HTTP status codes as follows:
```
tulshi, line 1446: There is no description of what the success response is to a deletion request. Please add.

openid-sse-framework-1_0.txt
```diff
@@ -1504,21 +1617,182 @@ Tulshibagwale, et al.   Expires December 10, 2021              [Page 26]
...
+     +------+--------------------------------------------------------+
+     | 404  | if there is no Event Stream with the given "stream_id" |
+     |      | for this Event Receiver, or if the Subject specified   |
+     |      | is invalid or if the Receiver is not authorized to get |
+     |      | status for the specified Subject.                      |
+     +------+--------------------------------------------------------+
+
+                     Table 6: Read Stream Status Errors
+
+   Examples:
```
tulshi, line 1645: These examples apply to all API calls. Should we move them to a part of the API description that applies to all calls?

openid-sse-framework-1_0.txt
```diff
@@ -1504,21 +1617,182 @@ Tulshibagwale, et al.   Expires December 10, 2021              [Page 26]
...
+       Transmitter policy does not permit the Receiver to read the
+       status of that Subject, then the Transmitter MAY respond with a
+       404 error status in order to not reveal the policy decision.
+
+   5.  If the specified Subject is invalid then the Transmitter MUST
+       respond with a 404 error status.
+
+7.1.2.2.  Updating a Stream's Status
+
+   An Event Receiver updates the current status of a stream by making an
```
tulshi, line 1667: we should mark this for a later pull request, but this needs to be changed to a PATCH since it's an update.

openid-sse-framework-1_0.txt
```diff
@@ -1504,21 +1617,182 @@ Tulshibagwale, et al.   Expires December 10, 2021              [Page 26]
...
+   Figure 28: Example: Update Stream Status Request With Optional Fields
+
+   The following is a non-normative example response:
+
+   HTTP/1.1 200 OK
+   Content-Type: application/json
+   Cache-Control: no-store
+
+   {
+     "status": "paused"
```
tulshi, line 1750: Should the response contain the stream Id at a minimum, and perhaps the subject for which the status was updated?

openid-sse-framework-1_0.txt
```diff
@@ -1681,13 +1967,14 @@ Tulshibagwale, et al.   Expires December 10, 2021              [Page 30]
    Authorization: Bearer eyJ0b2tlbiI6ImV4YW1wbGUifQo=
 
    {
+     "stream_id": "f67e39a0a4d34d56b3aa1bc4cff0069f",
      "subject": {
        "format": "phone",
        "phone_number": "+1 206 555 0123"
```
tulshi, line 1973: We should lose the spaces in the phone number format, according to the IETF SecEvents Subject Identifier spec.

openid-sse-framework-1_0.txt
```diff
@@ -1884,19 +2206,9 @@ Tulshibagwale, et al.   Expires December 10, 2021              [Page 33]
...
-
-
-
-Tulshibagwale, et al.   Expires December 10, 2021              [Page 34]
-

-                          openid-sse-framework                 June 2021
-
+   subject
+      OPTIONAL.  Specifies the Subject Principal for whom the status has
```
tulshi, line 2210: If this is optional, then we should specify that omitting the subject from the event means that the event applies to all subjects in the stream.

openid-sse-framework-1_0.txt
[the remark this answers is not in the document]
FragLegs: Will be fixed by #35

openid-sse-framework-1_0.txt
```diff
@@ -557,39 +552,44 @@ Tulshibagwale, et al.   Expires December 10, 2021               [Page 9]
...
 
     Figure 9: Example: Transmitter Configuration Request (without path)
 
-   If the Issuer value contains a path component, any terminating "/"
-   MUST be removed before inserting "/.well-known/sse-configuration"
-   between the host component and the path component.  The Receiver
-   would make the following request to the Issuer
-   "https://tr.example.com/issuer1" to obtain its Configuration
-   information, since the Issuer contains a path component:
+   If the Issuer value contains a path component, any terminating / MUST
```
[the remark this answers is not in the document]
FragLegs: Will be fixed by #35

openid-sse-framework-1_0.txt
FragLegs: Yes. I think the examples sections all need a bit of rework. Some endpoints don't have examples, others (like this one) have examples that are too general. Since this was already in the spec, I'd like to leave it for a different pull request that is focused strictly on examples, if that's alright.

openid-sse-framework-1_0.txt
FragLegs: 👍 I'll make an issue for this and the examples comment above.

openid-sse-framework-1_0.txt
FragLegs: See #37

openid-sse-framework-1_0.txt
FragLegs: Good catch!

openid-sse-framework-1_0.txt
FragLegs: It is specified in the paragraphs above the normative examples, like all of the other endpoints. Although I should have made it a `204` instead of a `201`. Fixing that now.

openid-sse-framework-1_0.txt
```diff
@@ -1544,14 +1882,17 @@ Tulshibagwale, et al.   Expires December 10, 2021              [Page 27]
...
    On a successful response, the Event Transmitter responds with an
-   empty "200 OK" response.  The Event Transmitter MAY choose to
-   silently ignore the request, for example if the subject has
-   previously indicated to the transmitter that they do not want events
-   to be transmitted to the Event Receiver.  In this case, the
-   transmitter MAY return an empty "200 OK" response or an appropriate
-   error code.  See Security Considerations (Section 9).
+   empty 200 OK response.  The Event Transmitter MAY choose to silently
+   ignore the request, for example if the subject has previously
+   indicated to the transmitter that they do not want events to be
```
topherMarie, line 1887: Transmitter is typically capitalized in this doc.

topherMarie (approved):
I flagged a few typographical issues. As I am just now familiarizing myself with the spec, I'm not sure my approval should count for much.

openid-sse-framework-1_0.txt
```diff
@@ -772,434 +749,358 @@ Tulshibagwale, et al.   Expires December 10, 2021              [Page 13]
...
 
+7.1.1.2.  Reading a Stream's Configuration
 
+   An Event Receiver gets the current configuration of a stream by
+   making an HTTP GET request to the Configuration Endpoint.  On
+   receiving a valid request the Event Transmitter responds with a 200
+   OK response containing a JSON [RFC7159] representation of the
+   stream's configuration in the body.
 
+   The GET request MUST include the "stream_id" as a parameter in order
```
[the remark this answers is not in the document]
FragLegs: We already have a section of the spec devoted to backwards compatibility for Google, and they could use that to maintain the existing behavior. If we did make `stream_id` optional, then I would expect this function to return a list of all streams. That's not a bad idea, but it's not backwards compatible.

openid-sse-framework-1_0.txt
```diff
@@ -1448,47 +1397,211 @@ Tulshibagwale, et al.   Expires December 10, 2021              [Page 25]
...
+                Table 4: Replace Stream Configuration Errors
+
+7.1.1.5.  Deleting a Stream
+
+   An Event Receiver deletes a stream by making an HTTP DELETE request
+   to the Configuration Endpoint.  On receiving a request the Event
+   Transmitter responds with a 201 OK response if the configuration was
+   successfully removed.
+
+   The DELETE request MUST include the "stream_id" as a parameter in
```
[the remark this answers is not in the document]
FragLegs: Same comment as above regarding backwards compatibility, except that an optional `stream_id` feels dangerous here, since it would delete all streams. That would be disastrous if you forgot to include the parameter. Since deleting streams should be a fairly rare event, I would lean towards making the caller always specify which stream to delete.

openid-sse-framework-1_0.txt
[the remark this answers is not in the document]
FragLegs: I have made `stream_id` optional and described what should happen in the case of no streams, one stream, and many streams. I think it is probably wise to provide Receivers with a way to get all of the stream in case they lose track of stream ids on their end.

openid-sse-framework-1_0.txt
FragLegs: Added

openid-sse-framework-1_0.xml
```diff
@@ -852,19 +633,125 @@ Cache-Control: no-store
...
+          <section anchor="creating-a-stream" title="Creating a Stream">
+            <t>In order to communicate events from a Transmitter to a Receiver, a Receiver
+              MUST first create an Event Stream. An Event Receiver creates a stream by making an HTTP POST
               request to the Configuration Endpoint. On receiving a valid request the Event
               Transmitter responds with a <spanx style="verb">200 OK</spanx> response containing a <xref target="RFC7159">JSON</xref>
               representation of the stream’s configuration in the body.</t>
 
-            <t>The following is a non-normative example request to read an Event Stream’s
-              configuration:</t>
+            <t>The HTTP POST request MAY contain the read/write values of the <xref target="stream-config">Stream Configuration</xref>
```
timcappalli, line 643: "read/write values" may be confusing. Since this is a create w/ POST, can we just say "values"?

openid-sse-framework-1_0.xml
```diff
@@ -887,7 +775,342 @@ Cache-Control: no-store
...
+
+            <t>Errors are signaled with HTTP status codes as follows:</t>
+
+            <texttable title="Read Stream Configuration Errors" anchor="tabreadconfig">
+              <ttcol align="left">Code</ttcol>
+              <ttcol align="left">Description</ttcol>
+              <c>401</c>
+              <c>if authorization failed or it is missing</c>
+              <c>403</c>
+              <c>if the Event Receiver is not allowed to read the stream configuration</c>
```
timcappalli, line 941: s/allowed/authorized

openid-sse-framework-1_0.xml
```diff
@@ -887,7 +775,342 @@ Cache-Control: no-store
...
+]]></artwork>
+            </figure>
+
+            <t>Errors are signaled with HTTP status codes as follows:</t>
+
+            <texttable title="Read Stream Configuration Errors" anchor="tabreadconfig">
+              <ttcol align="left">Code</ttcol>
+              <ttcol align="left">Description</ttcol>
+              <c>401</c>
+              <c>if authorization failed or it is missing</c>
```
timcappalli, line 939: "if credentials are missing or invalid"

openid-sse-framework-1_0.xml
```diff
@@ -852,19 +633,125 @@ Cache-Control: no-store
...
             <xref target="OAUTH-DISCOVERY"/>. This would allow other specs to add to the stream configuration.</t>
 
-          <section anchor="reading-a-streams-configuration"
-            title="Reading a Stream’s Configuration">
-            <t>An Event Receiver gets the current configuration of a stream by making an HTTP GET
+          <section anchor="creating-a-stream" title="Creating a Stream">
+            <t>In order to communicate events from a Transmitter to a Receiver, a Receiver
+              MUST first create an Event Stream. An Event Receiver creates a stream by making an HTTP POST
               request to the Configuration Endpoint. On receiving a valid request the Event
               Transmitter responds with a <spanx style="verb">200 OK</spanx> response containing a <xref target="RFC7159">JSON</xref>
```
timcappalli, line 640: Should this be 201 Created instead?

openid-sse-framework-1_0.xml
```diff
@@ -887,7 +775,342 @@ Cache-Control: no-store
...
+            </texttable>
+          </section>
+
+          <section anchor="updating-a-streams-configuration"
+            title="Updating a Stream’s Configuration">
+            <t>An Event Receiver updates the current configuration of a stream by making an HTTP
+              PATCH request to the Configuration Endpoint. The PATCH body contains a <xref target="RFC7159">JSON</xref>
+              representation of the stream configuration properties to change. On receiving a valid request the Event
+              Transmitter responds with a <spanx style="verb">200 OK</spanx> response containing a <xref target="RFC7159">JSON</xref>
+              representation of the updated stream configuration in the body.</t>
```
timcappalli, line 953: should we say "entire updated stream configuration" or is that implied?

openid-sse-framework-1_0.xml
```diff
@@ -887,7 +775,342 @@ Cache-Control: no-store
...
+            <texttable title="Update Stream Configuration Errors" anchor="tabupdateconfig">
+              <ttcol align="left">Code</ttcol>
+              <ttcol align="left">Description</ttcol>
+              <c>202</c>
+              <c>if the update request has been accepted, but not processed. Receiver MAY try the
+                same request later in order to get processing result.</c>
+              <c>400</c>
+              <c>if the request body cannot be parsed, a read-only property is incorrect, or if the request is otherwise invalid</c>
+              <c>401</c>
+              <c>if authorization failed or it is missing</c>
```
timcappalli, line 1036: "if credentials are missing or invalid"

openid-sse-framework-1_0.xml
```diff
@@ -904,157 +1127,308 @@ Cache-Control: no-store
...
+              <ttcol align="left">Description</ttcol>
+              <c>202</c>
+              <c>if the replace request has been accepted, but not processed. Receiver MAY try the
+                same request later in order to get processing result.</c>
+              <c>400</c>
+              <c>if the request body cannot be parsed, a read-only property is incorrect, or if the request is otherwise invalid</c>
+              <c>401</c>
+              <c>if authorization failed or it is missing</c>
+              <c>403</c>
+              <c>if the Event Receiver is not allowed to replace the stream configuration</c>
```
timcappalli, line 1146: s/allowed/authorized

timcappalli (commented):
Some minor stuff relating to errors and status codes

openid-sse-framework-1_0.xml
tulshi: +1. It will be better to make that explicit.

tulshi (approved):
minor comments, but LGTM otherwise.

openid-sse-framework-1_0.xml
FragLegs: I have always found the Read-Only vs Read-Write distinction of these values to be confusing. Would it make more sense to call them Transmitter-Supplied and Client-Supplied values?

openid-sse-framework-1_0.xml
```diff
@@ -904,157 +1127,308 @@ Cache-Control: no-store
...
+              <c>if the Event Receiver is not allowed to read the stream status</c>
               <c>404</c>
-              <c>if there is no Event Stream configured for this Event Receiver</c>
+              <c>if there is no Event Stream with the given "stream_id" for this Event Receiver, or if the Subject
+                specified is invalid or if the Receiver is not authorized to get status for the
+                specified Subject.</c>
             </texttable>
+            <t>Examples:
+              <list style="numbers">
+                <t>If a Receiver makes a request with an invalid OAuth token, then the Transmitter
```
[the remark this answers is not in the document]
FragLegs: If you are referring specifically to the Examples section, please see Issue #37. If you mean the definitions of the errors, then yes I do think they should be consistent. I would prefer that this PR leave them all the way they are currently written (so they are consistent) and we create an issue to update them based on your other comments:

"if credentials are missing or invalid"
s/allowed/authorized

openid-sse-framework-1_0.xml
FragLegs: I'd prefer to leave these consistent with the other error definitions in the spec. We can create a new issue to discuss changing all of them at once.

openid-sse-framework-1_0.xml
FragLegs: I'd prefer to leave these consistent with the other error definitions in the spec. We can create a new issue to discuss changing all of them at once.

openid-sse-framework-1_0.xml
FragLegs: 👍

openid-sse-framework-1_0.xml
FragLegs: I'd prefer to leave these consistent with the other error definitions in the spec. We can create a new issue to discuss changing all of them at once.

openid-sse-framework-1_0.xml
```diff
@@ -887,7 +775,342 @@ Cache-Control: no-store
...
+              <ttcol align="left">Description</ttcol>
+              <c>202</c>
+              <c>if the update request has been accepted, but not processed. Receiver MAY try the
+                same request later in order to get processing result.</c>
+              <c>400</c>
+              <c>if the request body cannot be parsed, a read-only property is incorrect, or if the request is otherwise invalid</c>
+              <c>401</c>
+              <c>if authorization failed or it is missing</c>
+              <c>403</c>
+              <c>if the Event Receiver is not allowed to update the stream configuration</c>
```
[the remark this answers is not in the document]
FragLegs: I'd prefer to leave these consistent with the other error definitions in the spec. We can create a new issue to discuss changing all of them at once.

openid-sse-framework-1_0.xml
```diff
@@ -904,157 +1127,308 @@ Cache-Control: no-store
...
+            <texttable title="Replace Stream Configuration Errors" anchor="tabreplaceconfig">
+              <ttcol align="left">Code</ttcol>
+              <ttcol align="left">Description</ttcol>
+              <c>202</c>
+              <c>if the replace request has been accepted, but not processed. Receiver MAY try the
+                same request later in order to get processing result.</c>
+              <c>400</c>
+              <c>if the request body cannot be parsed, a read-only property is incorrect, or if the request is otherwise invalid</c>
+              <c>401</c>
+              <c>if authorization failed or it is missing</c>
```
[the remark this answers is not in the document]
FragLegs: I'd prefer to leave these consistent with the other error definitions in the spec. We can create a new issue to discuss changing all of them at once.

openid-sse-framework-1_0.xml
FragLegs: I'd prefer to leave these consistent with the other error definitions in the spec. We can create a new issue to discuss changing all of them at once.

openid-sse-framework-1_0.xml
```diff
@@ -904,157 +1127,308 @@ Cache-Control: no-store
...
+]]></artwork>
+            </figure>
+
+            <t>Errors are signaled with HTTP status codes as follows:</t>
+
+            <texttable title="Delete Stream Errors" anchor="tabdeletestream">
+              <ttcol align="left">Code</ttcol>
+              <ttcol align="left">Description</ttcol>
+              <c>401</c>
+              <c>if authorization failed or it is missing</c>
```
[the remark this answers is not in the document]
FragLegs: I'd prefer to leave these consistent with the other error definitions in the spec. We can create a new issue to discuss changing all of them at once.

openid-sse-framework-1_0.xml
```diff
@@ -904,157 +1127,308 @@ Cache-Control: no-store
...
+              <c>403</c>
+              <c>if the Event Receiver is not allowed to delete the stream</c>
```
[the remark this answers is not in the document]
FragLegs: I'd prefer to leave these consistent with the other error definitions in the spec. We can create a new issue to discuss changing all of them at once.

openid-sse-framework-1_0.xml
timcappalli: Changes in [0f4f772](https://github.com/openid/sse/pull/9/commits/0f4f772db46c14489dac14dd2d2717be3f6c3ff1) look good to me!

timcappalli (approved):
LGTM!

PR #10, opened 2022-06-06
It's unnecessary to specify the character as UTF-8 in all examples

adeinega:
It's a default encoding for JSON. Furthermore, the specification tells us

> A successful response MUST use the 200 OK HTTP status code and return a JSON object using the "application/json" content.


timcappalli (approved):
lgtm

FragLegs (changes requested):
Needs to be copied to the txt and html files as well. See README for running the converter

adeinega:
@FragLegs, done.

FragLegs (approved):

PR #11, opened 2022-06-06
Pragma is an obsolete HTTP header

adeinega:
See https://bitbucket.org/openid/connect/issues/1483/directive-pragma-no-cache for more details.


FragLegs (changes requested):
Needs to be copied to the txt and html files as well. See README for running the converter

adeinega:
@FragLegs, done.

FragLegs (approved):

PR #31, opened 2022-09-11
Create namespace

tulshi:


openid-caep-specification-1_0.md
```diff
@@ -500,37 +539,38 @@ nested `reason_admin` and/or `reason_user` claims made in {{optional-event-claim
...
-  - `nist-aal1`
-  - `nist-aal2`
-  - `nist-aal3`
+  - `RFC8176`: The assurance level values are from the {{RFC8176}} specification
+  - `RFC6711`: The assurance level values are from the {{RFC6711}} specification
+  - `ISO-IEC-29115`: The assurance level values are from the {{ISO-IEC-29115}} specification
+  - `NIST-IAL`: The assurance level values are from the {{NIST-IDPROOF}} specification
+  - `NIST-AAL`: The assurance level values are from the {{NIST-AUTH}} specification
+  - `NIST-FAL`: The assurance level values are from the {{NIST-FED}} specification
+  - Any other value that is an aliaas for a custom namespace agreed between the Transmitter and the Receiver
```
FragLegs, line 552:
```suggestion
  - Any other value that is an alias for a custom namespace agreed between the Transmitter and the Receiver
```

FragLegs (approved):
One small typo that needs to be fixed in all 4 files, but otherwise looks good.

openid-caep-specification-1_0.txt
```diff
@@ -703,6 +713,45 @@ Cappalli & Tulshibagwale     Standards Track                   [Page 12]
...
+       "iat": 1615305159,
+       "aud": "https://sp.example2.net/caep",
+       "events": {
+           "https://schemas.openid.net/secevent/caep/event-type/assurance-level-change": {
+               "subject": {
+                   "format": "iss_sub",
+                   "iss": "https://idp.example.com/3456789/",
+                   "sub": "jane.smith@example.com"
+               },
+               "namespace": "Retinal Scan",
```
timcappalli, line 745: We should use an example that is from an actual registry since that is the primary use case.

Suggest we use NIST-AAL
```
...
"namespace": "NIST-IAL",
"current_level": "IAL2",
...
```

openid-caep-specification-1_0.txt
```diff
@@ -630,55 +630,64 @@ Cappalli & Tulshibagwale     Standards Track                   [Page 11]
 
 3.4.1.  Event-Specific Claims
 
-   current_level  REQUIRED, JSON string: the current NIST Authenticator
-      Assurance Level (AAL) as defined in [SP800-63R3]
+   namespace:  REQUIRED, JSON string: the namespace of the values in the
```
timcappalli, line 633: JSON is technically case sensitive. Should we say that this string is case-insensitive?

openid-caep-specification-1_0.txt
```diff
@@ -630,55 +630,64 @@ Cappalli & Tulshibagwale     Standards Track                   [Page 11]
...
-      This MUST be one of the following strings:
+      *  NIST-FAL: The assurance level values are from the [NIST-FED]
+         specification
 
-      *  nist-aal1
+      *  Any other value that is an alias for a custom namespace agreed
+         between the Transmitter and the Receiver
 
-      *  nist-aal2
+   current_level  REQUIRED, JSON string: The current assurance level, as
```
timcappalli, line 658: JSON is technically case sensitive. Should we say that this string is case-insensitive?

openid-caep-specification-1_0.txt
```diff
@@ -630,55 +630,64 @@ Cappalli & Tulshibagwale     Standards Track                   [Page 11]
...
 
-      *  nist-aal3
+   previous_level  OPTIONAL, JSON string: the previous assurance level,
+      as defined in the specified namespace If the Transmitter omits
+      this value, the Receiver MUST assume that the previous assurance
+      level is unknown to the Transmitter
 
-   change_direction  REQUIRED, JSON string: the Authenticator Assurance
-      Level increased or decreased
+   change_direction  OPTIONAL, JSON string: the Authenticator Assurance
```
timcappalli, line 666: JSON is technically case sensitive. Should we say that this string is case-insensitive?

timcappalli, line 666: I think we need to update the text here as it is still tied to the original authenticator assurance use case

openid-caep-specification-1_0.txt
tulshi: This example is to show a "custom" assurance level, not one defined in the spec. NIST-IAL and NIST-AAL are defined in the spec.

openid-caep-specification-1_0.txt
tulshi: It might be good to add it as a top-level guidance for anything defined in the specification

openid-caep-specification-1_0.txt
tulshi: It might still apply because if a Transmitter has specified a Previous Level, then the Transmitter *should* specify how the new level relates to the previous one.

openid-caep-specification-1_0.md
```diff
@@ -144,6 +144,9 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 document are to be interpreted as described in BCP 14 {{RFC2119}} {{RFC8174}} 
 when, and only when, they appear in all capitals, as shown here.
 
+# Case Insensitivity
```
tulshi, line 147: @timcappalli does this address your previous comment re: case insensitivity?

timcappalli (approved):
lgtm

PR #35, opened 2022-11-21
Add special handling to the Makefile to re-insert quotes around <spanx style='verb'> sections when converting to text format

FragLegs:
- Adds a python script for converting `<spanx style="verb"></spanx> sections to quoted sections for the *.txt files
- Updates the Makefile to use the new python script to generate a temporary XML file from which to generate the TXT file (then cleans up the temp XML file)
- Generates the TXT files for the three specs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,9 @@ html:   $(HTML)
 	xml2rfc --html $^
 
 %.txt:	%.xml
-	xml2rfc $^ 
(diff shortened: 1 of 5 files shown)


FragLegs:
I don't know why this output changes the expiration dates to "Standards Track" or why it changes some of the page breaks. We may want to consider shipping a canonical version of the conversion scripts (a dockerfile probably) with this repo to avoid conflicting edits in auto-generated text.

PR #41, opened 2023-01-24
renamed files to be called sharedsignals instead of SSE. No change to…

tulshi:
… the content


FragLegs (approved):
I used `meld` to confirm that the files are identical. Thanks!

timcappalli (approved):

Issue #28, opened 2022-08-19
[SSE] Delivery method references


timcappalli:
Phil Hunt has asked why the delivery methods in SSE are referenced by an OpenID Schema URI instead of direct references to the RFCs.

timcappalli:
These should be updated to:
* `urn:ietf:rfc:8935`
* `urn:ietf:rfc:8936`

PR #49, opened 2023-03-21
#28 update delivery method references to URNs

timcappalli:
- replace all references to `https://schemas.openid.net/secevent/risc/delivery-method/push` with `urn:ietf:rfc:8935`
- replace all references to `https://schemas.openid.net/secevent/risc/delivery-method/poll` with `urn:ietf:rfc:8936`


tulshi (approved):

FragLegs:
Is there any way we could make it clear in our spec that these two URNs represent push/poll for readers who are not well-versed with those standards?

tulshi (approved):

FragLegs (approved):

Issue #46, opened 2023-03-03
Create Stream description lacks option for "Stream already exists"


tulshi:
If a Transmitter wanted to support only one stream between itself and a specific Receiver, then the Transmitter should have a way to respond to a Receiver that has already created a stream with a status 303 (see other) or some such way of indicating that a stream already exists.

ptravers:
I would suggest returning 200 with the previously created entity. then return 201 if the stream did not already exist when create is called

timcappalli:
There is precedent for @ptravers recommendation in the [SCIM core protocol](https://www.rfc-editor.org/rfc/rfc7644.html#section-3.4.1)

>  If the resource exists, the server responds with HTTP status code 200 (OK) and includes the result in the body of the response.

on 2023-03-21 call, consensus to follow this approach.

PR #50, opened 2023-03-22
#46 add stream exists behavior

timcappalli:
Issue #46: add stream exists behavior


openid-sharedsignals-framework-1_0.md
```diff
@@ -829,6 +829,10 @@ an HTTP POST request to the Configuration Endpoint. On receiving a valid request
 the Event Transmitter responds with a "201 Created" response containing a
 [JSON][RFC7159] representation of the stream’s configuration in the body.
 
+If the stream already exists, the Event Transmitter responds with HTTP status code
```
FragLegs, line 832:
```suggestion
If a stream already exists, the expected behavior is for the Transmitter to create a new stream with a different ID. If the Transmitter does not want to allow multiple streams per receiver, it MAY respond with HTTP status code
```

openid-sharedsignals-framework-1_0.md
```diff
@@ -829,6 +829,10 @@ an HTTP POST request to the Configuration Endpoint. On receiving a valid request
...
+"200 OK" and includes a [JSON][RFC7159] representation of the stream’s configuration
```
FragLegs, line 833:
```suggestion
"200 OK" and includes a [JSON][RFC7159] representation of the existing stream’s configuration
```

FragLegs (changes requested):
I think we should be clear about the expected behavior and that only allowing one stream per receiver is the uncommon case.

openid-sharedsignals-framework-1_0.md
tulshi: I agree with changing "the stream" to "a stream". However, instead of using "expected behavior is..." we should use the verbs "SHOULD" or "MUST" as appropriate. I feel having a "MUST" here will help interoperability. I suggest the following text: "If a stream already exists, and the Transmitter allows multiple streams with the same Receiver, then the Transmitter MUST respond with a new stream ID. If the Transmitter does not allow multiple streams with the same Receiver, it MUST respond with the HTTP status code 200 OK and include the representation of the existing stream's configuration".

timcappalli: Addressed in [9497085](https://github.com/openid/sharedsignals/pull/50/commits/94970859e67adfb2248a79703986d356fdc2cb04)

openid-sharedsignals-framework-1_0.md
timcappalli: Addressed in [9497085](https://github.com/openid/sharedsignals/pull/50/commits/94970859e67adfb2248a79703986d356fdc2cb04)

openid-sharedsignals-framework-1_0.md
```diff
@@ -829,6 +829,12 @@ an HTTP POST request to the Configuration Endpoint. On receiving a valid request
 the Event Transmitter responds with a "201 Created" response containing a
 [JSON][RFC7159] representation of the stream’s configuration in the body.
 
+If a stream already exists, and the Transmitter allows multiple streams with the
+same Receiver, the Event Transmitter MUST respond with a new stream ID. If the
+Transmitter does not allow multiple streams with the same Receiver, it MUST respond
+respond with HTTP status code "200 OK" and include a [JSON][RFC7159] representation
```
FragLegs, line 835: What should happen if the Transmitter does not allow multiple streams, a stream already exists, and the details of that stream are different from the values in this POST request?

FragLegs: I think the last sentence should be changed to say

If the Transmitter does not allow multiple streams with the same Receiver, it MUST respond with a "409 - Conflict" error code. The Receiver MAY then GET the existing stream configuration and, if desired, use PATCH or PUT to update or replace the existing stream configuration.

FragLegs (changes requested):

openid-sharedsignals-framework-1_0.md
appsdesh: From [the description of 409](https://www.rfc-editor.org/rfc/rfc9110#status.409) it looks to be remediable which is not the case here. I think we should stay with 200 OK as return the existing stream config

Adding a new status code would require the existing implementation to change the error handling when this draft is approved. 

Also, may break idempotency expectations with different outcome for same request

FragLegs: > Also, may break idempotency expectations with different outcome for same request

Good point! Ok, then we need to address the case where a Receiver tries to "create" a stream a second time with different information. What should happen? An error? A silent update?

appsdesh: > What should happen? An error? A silent update?

Another option is to ignore the input and respond with the existing stream. The stream ID would indicate that the transmitter did not create a new stream

timcappalli: 2023-08-01 call: consensus on 409

tulshi:
Not sure why the HTML and TXT files are showing up. Do you need to rebase?

openid-sharedsignals-framework-1_0.md
```diff
@@ -829,6 +829,13 @@ an HTTP POST request to the Configuration Endpoint. On receiving a valid request
 the Event Transmitter responds with a "201 Created" response containing a
 [JSON][RFC7159] representation of the stream’s configuration in the body.
 
+If a stream already exists, and the Transmitter allows multiple streams with the
+same Receiver, the Event Transmitter MUST respond with a new stream ID. If the
+Transmitter does not allow multiple streams with the same Receiver, it MUST respond
+respond with HTTP status code "409 Conflict". The Receiver MAY then GET the
+existing stream configuration and, if desired, use PATCH or PUT to update or
+replace the existing stream configuration.
+
```
appsdesh, line 838: Please add 409 in the error table below

appsdesh (changes requested):

independentid:
409 may not be appropriate. If you look at the definition the problem is that the client should try again...

https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.9

403 seems appropriate because it says retrying won't fix the issue.

timcappalli:
> 403 seems appropriate because it says retrying won't fix the issue.

We spent almost an hour debating this one yesterday. I don't think 403 is appropriate either as it talks about authorization.

"...but refuses to authorize it."

We landed on 409 because the receiver can ultimately resolve this by changing their request.

independentid:
In what way can the receiver change their request?  It is authorization because the server has a policy of allowing only a single stream.  403 means an administrator has to take action to resolve (what is needed). 409 literally means try the request again immediately. In my experience 409 has always mean execute a retry.  PhilOn Aug 2, 2023, at 6:51 AM, Tim Cappalli ***@***.***> wrote:﻿

403 seems appropriate because it says retrying won't fix the issue.

We spent almost an hour debating this one yesterday. I don't think 403 is appropriate either as it talks about authorization.
"...but refuses to authorize it."
We landed on 409 because the receiver can ultimately resolve this by changing their request.

independentid:
My regrets on not making the call. It seems I have to update my calendar as all my OpenID meetings have disappeared.

Reading from the minutes of last tuesday, I got the impression that the group was trying to assign high-level meaning to the response.  When dealing with HTTP Status messages you have to always think about protocol meaning. In RFC7231, the client is allowed to immediately re-try the request (e.g. beause another client PUT or POSTED at the same time).  

For me the issue is accademic.  If i2goSignals was only going to allow 1 stream it would return 403 because the client is only authorized to have 1 stream.  From a protocol perspective, whether the reason is technical or policy, the protocol response should be the same - because we should want the requestor to stop trying.

Answering Tim’s comment, if an endpoint only supports 1 stream, why would it implement SSF?  Implementing stream management seems like a burden.

i2scim.io will only support 1 incoming stream and 1 outgoing stream.  It has no stream management because it uses SET PUSH to deliver and SET POLL to receive.  Stream management is what i2scim.io calls as an HTTP client  to set itself up.  


Phillip Hunt
***@***.***

tulshi (approved):

FragLegs (approved):
Looks good!

Issue #45, opened 2023-03-03
Before a Receiver creates a stream, there is no way for it to discover which events a Transmitter supports for it


tulshi:
Section 7.1.1.2 specifies that before a stream is created the response sent by the Transmitter is empty. It should be the list of events supported by the Transmitter for that Receiver.

FragLegs:
Good catch! I wonder if we should move `events_supported` to the Transmitter Configuration instead of the Stream Configuration. It's more a property of the transmitter than the stream - i.e. it won't change from stream to stream. And it would get us around having to return a somewhat kludgy Stream Configuration response when no streams have been created.

tulshi:
It could be receiver specific though. A Transmitter may want to support one set of events for one Receiver, and another set of events for another Receiver.

tulshi:
See notes https://hackmd.io/@oidf-wg-sse/wg-meeting-20230606, only change will be to the description of "events requested" and "events delivered" in the stream configuration description. This functionality (of discovering events supported before creating a stream) need not be added to the spec.

PR #69, opened 2023-06-06
Poll transmitter specified url

### LINKED ISSUE #36: Should POLL URL be specified in the "delivery methods supported"?
From @atultulshi in a comment on #9 

> In the Transmitter Configuration Metadata, should we modify the "delivery methods supported" to include the POLL URL if the method is POLL?

tulshi:
Closes #36 (previous commits incorrectly refer to #45)

### DIFF
--- a/openid-sharedsignals-framework-1_0.html
+++ b/openid-sharedsignals-framework-1_0.html
@@ -2348,7 +2348,7 @@ following properties:<a href="#section-7.1.1-1" class="pilcrow">¶</a></p>
           </ul>
 <ul class="normal ulEmpty">
 <li class="normal ulEmpty" id="section-7.1.1-16.1">
-              <p id="section-7.1.1-16.1.1">delivery_method<a href="#section-7.1.1-16.1.1" class="pilcrow">¶</a></p>
+              <p id="section-7.1.1-16.1.1">method<a href="#section-7.1.1-16.1.1" class="pilcrow">¶</a></p>
 </li>
           </ul>
 <ul class="normal ulEmpty">
@@ -2372,8 +2372,8 @@ following properties:<a href="#section-7.1.1-1" class="pilcrow">¶</a></p>
               <ul class="normal ulEmpty">
 <li class="normal ulEmpty" id="section-7.1.1-19.1.1.1">
                   <p id="section-7.1.1-19.1.1.1.1">The location at which the push or poll delivery will take place. If the
-    <code>delivery_method</code> value is "urn:ietf:rfc:8935" (push), then this value MUST
-    be supplied by the Receiver.  If the <code>delivery_method</code> value is
+    <code>method</code> value is "urn:ietf:rfc:8935" (push), then this value MUST
+    be supplied by the Receiver.  If the <code>method</code> value is
     "urn:ietf:rfc:8936" (poll), then this value MUST be supplied by the
     Transmitter.<a href="#section-7.1.1-19.1.1.1.1" class="pilcrow">¶</a></p>
 </li>
@@ -2426,9 +2426,9 @@ supplied by the Transmitter.<a href="#section-7.1.1.1-3.2" class="pilcrow">¶</a
 </li>
             </ul>
 <p id="section-7.1.1.1-4">If the request does not contain the <code>delivery</code> property, then the Transmitter
-MUST assume that the <code>delivery_method</code> is "urn:ietf:rfc:8936" (poll). The
+MUST assume that the <code>method</code> is "urn:ietf:rfc:8936" (poll). The
 Transmitter MUST include a <code>delivery</code> property in the response with this
-delivery_method and a <code>url</code> property.<a href="#section-7.1.1.1-4" class="pilcrow">¶</a></p>
+<code>method</code> property and a <code>url</code> property.<a href="#section-7.1.1.1-4" class="pilcrow">¶</a></p>
 <p id="section-7.1.1.1-5">The following is a non-normative example request to create an Event Stream:<a href="#section-7.1.1.1-5" class="pilcrow">¶</a></p>
 <span id="name-example-create-event-stream"></span><div id="figcreatestreamreq">
 <figure id="figure-14">
@@ -2440,7 +2440,7 @@ Authorization: Bearer eyJ0b2tlbiI6ImV4YW1wbGUifQo=
 
 {
   "delivery": {
-    "delivery_method": "urn:ietf:rfc:8935",
+    "method": "urn:ietf:rfc:8935",
     "url": "https://receiver.example.com/events"
   },
   "events_requested": [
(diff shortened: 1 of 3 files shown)


openid-sharedsignals-framework-1_0.md
```diff
@@ -798,10 +798,24 @@ events_delivered
...
+  parameters for the SET delivery method. The actual delivery method is
+  identified by the special key "method" with the value being a URI as defined
+  in {{delivery-meta}}. The value of the "delivery" field contains two
+  sub-fields: 
+  
+>   delivery_method
+
+> > **Receiver-Supplied**, the specific delivery method to be used. This can be
+    any one of "push" (as defined in {{delivery-push}}) or "poll" (as defined
+    in {{delivery-poll}}), but not both.
```
appsdesh, line 811: Should we align on the convention and use URN instead of "push" vs "pull" as per #49  changes?

openid-sharedsignals-framework-1_0.md
```diff
@@ -798,10 +798,24 @@ events_delivered
...
-  specifying configuration parameters for the SET delivery method.  The actual
-  delivery method is identified by the special key "method" with the value being
-  a URI as defined in {{delivery-meta}}.
+> A JSON object containing a set of name/value pairs specifying configuration
+  parameters for the SET delivery method. The actual delivery method is
+  identified by the special key "method" with the value being a URI as defined
+  in {{delivery-meta}}. The value of the "delivery" field contains two
+  sub-fields: 
+  
+>   delivery_method
```
appsdesh, line 807: `delivery_` is redundant as it's part of the `delivery` object. Also changing the attribute name would break the backwards compatibility.

I suggest we maintain `method` as an inner attribute name

timcappalli: Without understanding the goal of this change, I agree with @appsdesh.

openid-sharedsignals-framework-1_0.md
```diff
@@ -832,24 +846,10 @@ the Event Transmitter responds with a "201 Created" response containing a
...
-  delivery method is identified by the special key "method" with the value
-  being a URI as defined in {{delivery-meta}}.
-
-format
-
-> **Receiver-Supplied**, The Subject Identifier Format that the Receiver wants
-  for the events. If not set then the Transmitter might decide to use a type
-  that discloses more information than necessary.
+* `events_requested`
+* `delivery` : Note that in the case of the POLL method, the `url` value is
```
FragLegs, line 850: Since Receivers MAY include these values, we should say that if the `delivery` value is not included in the stream configuration request, a Transmitter MUST respond with polling as the delivery method.

FragLegs (commented):
I think you originally intended for this to be a documentation-only change. It might be easier if that's all it was and we left the renaming of `method` and `endpoint_url` for a different pull request.

openid-sharedsignals-framework-1_0.md
tulshi: added

openid-sharedsignals-framework-1_0.md
tulshi: The current spec uses "delivery_method" and not "method", although "method" is mentioned in the profile section. The intent of this PR is to clarify that in case of POLL, the "url" property should be supplied by the Transmitter. I am therefore not changing the name from "delivery_method" to "method" in this PR. We can have a separate PR to cover that.

tulshi: Actually, it looks like the examples give the wrong name "delivery_method", instead of the correct one "method". I will modify the PR to use "method". Stay tuned.

FragLegs (approved):

openid-sharedsignals-framework-1_0.md
```diff
@@ -800,10 +800,25 @@ events_delivered
 
 delivery
 
-> **Receiver-Supplied**, A JSON object containing a set of name/value pairs
-  specifying configuration parameters for the SET delivery method.  The actual
-  delivery method is identified by the special key "method" with the value being
-  a URI as defined in {{delivery-meta}}.
+> A JSON object containing a set of name/value pairs specifying configuration
+  parameters for the SET delivery method. The actual delivery method is
+  identified by the special key "method" with the value being a URI as defined
```
timcappalli, line 805: minor nit. Why is this key "special"?

openid-sharedsignals-framework-1_0.html
```diff
@@ -2378,43 +2413,34 @@ <h5 id="name-creating-a-stream">
...
+              <div class="lang-http sourcecode" id="section-7.1.1.1-6.1">
 <pre>
 POST /ssf/stream HTTP/1.1
 Host: transmitter.example.com
 Authorization: Bearer eyJ0b2tlbiI6ImV4YW1wbGUifQo=
 
 {
   "delivery": {
-    "delivery_method": "urn:ietf:rfc:8935",
+    "method": "urn:ietf:rfc:8935",
```
appsdesh, line 2443: I looked at the already approved spec and it appears that the spec uses `delivery_method` key.
```
"delivery": {
    "delivery_method":
      "https://schemas.openid.net/secevent/risc/delivery-method/push",
      "url": "https://receiver.example.com/events"
  },
```
https://openid.net/specs/openid-sse-framework-1_0-ID1.html#rfc.section.7.1.2.2

Though I would have liked the `method` over `delivery_method` (as described earlier), that would break the backwards compatibility of transmitter's APIs.

FragLegs: The currently approved spec uses `method` in the spec (see [sections 11.2.1.1. and 11.2.1.2](https://openid.net/specs/openid-sse-framework-1_0.html#rfc.section.11.2.1.1)), and mistakenly uses `delivery_method` in the examples.

appsdesh: Thank you for clarifying it :)

appsdesh: @FragLegs @tulshi noticed discrepancy around `endpoint_url` vs `url`. The spec mentions `endpoint_url` (see [sections 11.2.1.1. and 11.2.1.2](https://openid.net/specs/openid-sse-framework-1_0.html#rfc.section.11.2.1.1)) 

And examples show `url`. Could we align on this either in this PR or a brand new?

openid-sharedsignals-framework-1_0.md
```diff
@@ -800,10 +800,25 @@ events_delivered
...
+  in {{delivery-meta}}. The value of the "delivery" field contains two
+  sub-fields: 
+  
+>   method
+
+> > **Receiver-Supplied**, the specific delivery method to be used. This can be
+    any one of "urn:ietf:rfc:8935" (push) or "urn:ietf:rfc:8936" (poll), but
+    not both.
```
adeinega, line 813: The method field is `String`, not sure how it's possible to specify them both :)

If this specification wants to be flexible then it might make sense to allow defining other values (delivery methods).

tulshi: Does not seem ambiguous. Please reopen issue if you think this is not resolved.

timcappalli (approved):
lgtm

PR #70, opened 2023-06-20
Fixes #60 - are subjecst required

### LINKED ISSUE #60: Are subjects required in SSF events?
The spec says that the `subject` claim is [REQUIRED for all SSF events](https://github.com/openid/sharedsignals/blob/main/openid-sharedsignals-framework-1_0.txt#L2716) but also makes the `subject` [OPTIONAL for StatusChanged events](https://github.com/openid/sharedsignals/blob/main/openid-sharedsignals-framework-1_0.txt#L2525) and doesn't mention `subject` at all [for the Verification event](https://github.com/openid/sharedsignals/blob/main/openid-sharedsignals-framework-1_0.txt#L2317).

We need to resolve this inconsistency.

tulshi:


openid-sharedsignals-framework-1_0.md
```diff
@@ -1718,6 +1718,16 @@ state
 > OPTIONAL An opaque value provided by the Event Receiver when the event is
   triggered. This is a nested attribute in the event payload.
 
+subject
+
+> REQUIRED. The value of the `subject` field in a Verification Event MUST always
+  be set to have a simple value of type `opaque`. The `id` of the value MUST be
+  the `stream_id` of the stream being verified. If the Transmitter does not
+  support multiple streams, this value MUST be set to "0"
```
FragLegs, line 1726: Doesn't the Transmitter still assign a stream ID even when it only supports one stream? If it doesn't then all of our other endpoints are broken.

openid-sharedsignals-framework-1_0.md
```diff
@@ -1802,6 +1812,10 @@ Event Receiver as a result of the above request:
   "iat": 1493856000,
   "events": {
     "https://schemas.openid.net/secevent/ssf/event-type/verification":{
+      "subject": {
+        "format": "opaque",
+        "id": "1234"
```
FragLegs, line 1817: For clarity, this should be the stream ID from the example request above

openid-sharedsignals-framework-1_0.md
```diff
@@ -1838,9 +1852,18 @@ reason
 
 subject
 
-> OPTIONAL. Specifies the Subject Principal for whom the status has been updated.
+> REQUIRED. Specifies the Subject Principal for whom the status has been updated.
   If this claim is not included, then the status change was applied to all
```
FragLegs, line 1856: It can't be REQUIRED and not included. This sentence should be edited.

openid-sharedsignals-framework-1_0.md
```diff
@@ -1838,9 +1852,18 @@ reason
...
-  subjects in the stream.
+  subjects in the stream. If the event applies to the entire stream, the value
+  of the `subject` field MUST be of format `opaque`, and its `id` value MUST be
+  the unique ID of the stream. If the Transmitter does not support multiple
+  streams, then the value of the `id` field MUST be the string "0".
```
FragLegs, line 1860: Same comment as above. Even when a Transmitter only supports a single stream, they must provide a stream ID

FragLegs (changes requested):

openid-sharedsignals-framework-1_0.md
tulshi: Changed this to "none" to avoid the idea that a single-stream Transmitter must also name their only stream.

FragLegs (approved):

timcappalli (approved):
lgtm

PR #81, opened 2023-07-16
Create azure-webapps-node.yml

gilbertalgordo:


timcappalli:
spam

gilbertalgordo:
Not spam I just not subscribe to azure cloud but on Ql

PR #82, opened 2023-07-17
added new sub_id related language. Modified existing language to refl…

tulshi:
…ect the use of the sub_id claim


independentid:
The format value is an IANA registered format to avoid conflicts. The term "complex" is probably too generic.

You may want to prefix it with the spec name (e.g. caep, risc, ssf) For example "caep_complex".

Registration info is in section [8.1 of the subject identifier draft](https://datatracker.ietf.org/doc/html/draft-ietf-secevent-subject-identifiers-18#name-security-event-identifier-f). 

Other than that looks good.

openid-sharedsignals-framework-1_0.md
```diff
@@ -240,10 +266,20 @@ Subject Principals are identified by Subject Members defined below.
 
 # Subject Members in SSF Events {#subject-ids}
 
-A member of type Subject in an SSF event MAY have any claim name. Each Subject Member MUST
-refer to exactly one Subject Principal.
+## Subject Members {#subject-members}
+A Subject Member of a SSF event describes a subject of the event. A top-level claim named `sub_id` SHOULD be used to describe the primary Subject Principal of the event.
```
FragLegs, line 270: Can we say MUST instead of SHOULD? We can make the `sub_id` claim required without messing with the CAEP or RISC `subject` claim.

openid-sharedsignals-framework-1_0.md
```diff
@@ -1373,7 +1434,7 @@ stream_id
 
 > REQUIRED. The stream whose status is being queried.
 
-subject
+sub_id
```
FragLegs, line 1437: I don't know if I like changing these arguments as part of this pull request. `subject` is clearer and there is no requirement that the argument name to this function matches the claim name in the SET that get produced by this Transmitter.

openid-sharedsignals-framework-1_0.md
```diff
@@ -1495,7 +1556,7 @@ status
 
 > REQUIRED. The new status of the Event Stream.
 
-subject
+sub_id
```
FragLegs, line 1559: Same comment as above

openid-sharedsignals-framework-1_0.md
```diff
@@ -1596,7 +1657,7 @@ stream_id
 
 > REQUIRED. The stream to which the subject is being added.
 
-subject
+sub_id
```
FragLegs, line 1660: Same comment as above

openid-sharedsignals-framework-1_0.md
```diff
@@ -1662,7 +1723,7 @@ stream_id
 
 > REQUIRED. The stream from which the subject is being removed.
 
-subject
+sub_id
```
FragLegs, line 1726: Same comment as above

openid-sharedsignals-framework-1_0.md
```diff
@@ -1732,11 +1793,9 @@ state
 > OPTIONAL An opaque value provided by the Event Receiver when the event is
   triggered. This is a nested attribute in the event payload.
 
-subject
+sub_id
```
FragLegs, line 1796: Same comment as above

FragLegs (changes requested):

FragLegs:
> The format value is an IANA registered format to avoid conflicts. The term "complex" is probably too generic.
> 
> You may want to prefix it with the spec name (e.g. caep, risc, ssf) For example "caep_complex".
> 
> Registration info is in section [8.1 of the subject identifier draft](https://datatracker.ietf.org/doc/html/draft-ietf-secevent-subject-identifiers-18#name-security-event-identifier-f).
> 

@independentid Can we separate that into another issue/pull request to avoid polluting the discussion of this one?

independentid:
Sure. Could be handled as a separate issue.

openid-sharedsignals-framework-1_0.md
tulshi: I think the concern was that if there are existing Transmitters that don't insert the sub_id claim, then they will become incompatible with the updated spec.

openid-sharedsignals-framework-1_0.md
tulshi: This is a field in the verification event. On the call on 7/18, we agreed that the events should contain "sub_id" claims, but the requests may, where appropriate, use the term "subject"

openid-sharedsignals-framework-1_0.md
```diff
@@ -1863,10 +1922,10 @@ reason
 > OPTIONAL. Provides a short description of why the Transmitter has updated the
   status.
 
-subject
+sub_id
```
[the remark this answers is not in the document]
tulshi: This too is an event field

openid-sharedsignals-framework-1_0.md
```diff
@@ -240,10 +266,20 @@ Subject Principals are identified by Subject Members defined below.
...
+A Subject Member of a SSF event describes a subject of the event. A top-level claim named `sub_id` SHOULD be used to describe the primary Subject Principal of the event.
+
+### Existing CAEP and RISC Events
+Event types already defiend in the CAEP ({{CAEP}}) and RISC ({{RISC}}) drafts MAY use a `subject` field within the `events` claim of the SSF eent to describe the primary Subject Principal of the event. SSF Transmitters SHOULD include the top-level `sub_id` claim even for these existing event types.
+
+### New Event Types
+New event types MUST use the top-level `sub_id` claim and MUST NOT use the `subject` field in the `events` claim to describe the primary Subject Principal.
 
-A Subject may be a "simple subject" or a "complex subject".
+### Additional Subject Members
```
appsdesh, line 278: Could you kindly indicate the location where the "Additional Subject Members" would be listed?

openid-sharedsignals-framework-1_0.md
```diff
@@ -1838,12 +1897,12 @@ Event Receiver as a result of the above request:
   "iss": "https://transmitter.example.com",
   "aud": "receiver.example.com",
   "iat": 1493856000,
+  "sub_id": {
+    "format": "opaque",
+    "id": "f67e39a0a4d34d56b3aa1bc4cff0069f"
+  },
   "events": {
     "https://schemas.openid.net/secevent/ssf/event-type/verification":{
-      "subject": {
```
appsdesh, line 1843: I think we said that "subject" in the event will be retained but called out as deprecated.

openid-sharedsignals-framework-1_0.md
```diff
@@ -1897,16 +1956,16 @@ subject
   "iss": "https://transmitter.example.com",
   "aud": "receiver.example.com",
   "iat": 1493856000,
-  "events": {
-    "https://schemas.openid.net/secevent/ssf/event-type/stream-updated": {
-      "subject": {
```
appsdesh, line 1902: I think we said that "subject" in the event will be retained but called out as deprecated.

openid-sharedsignals-framework-1_0.md
```diff
@@ -1924,12 +1983,12 @@ subject
   "iss": "https://transmitter.example.com",
   "aud": "receiver.example.com",
   "iat": 1493856000,
+  "sub_id": {
+    "format": "opaque",
+    "id" : "f67e39a0a4d34d56b3aa1bc4cff0069f"
+  },   
   "events": {
     "https://schemas.openid.net/secevent/ssf/event-type/stream-updated": {
-      "subject": {
```
appsdesh, line 1929: I think we said that "subject" in the event will be retained but called out as deprecated.

openid-sharedsignals-framework-1_0.md
FragLegs: The downside of this decision is that Receivers will forever have to look in two places to find the subject information and, because it is not required in either place, it would be possible (although unlikely) for a Transmitter to send a SET with no subject information.

This draft already breaks backwards compatibility. I think our goal should be no breakage _without a good reason_.

openid-sharedsignals-framework-1_0.md
FragLegs: I think it would be totally up to the authors of the new event types. For instance, I might create a "got married" event that looks like
```
{
  "date": "2009-11-23",
  "location": "New York, NY",
  "subject_1": "Avery Apple",
  "subject_2": "Blair Bonavue"
}
```

This section is just saying that it is perfectly fine for me to create new event types that include subject information, and that I am not restricted in the naming of those fields.

openid-sharedsignals-framework-1_0.md
FragLegs: Oh, you are right! We should move this field out of the Verification event schema, since it is supposed to be in the top level of the SET. We just need a note somewhere indicating how to specify the `sub_id` in the SET. This makes it look like the SET should be

```
{
  "jti": "123456",
  "iss": "https://transmitter.example.com",
  "aud": "receiver.example.com",
  "iat": 1493856000,
  "sub_id": {
    "format": "opaque",
    "id": "f67e39a0a4d34d56b3aa1bc4cff0069f"
  },
  "events": {
    "https://schemas.openid.net/secevent/ssf/event-type/verification":{
      "state": "VGhpcyBpcyBhbiBleGFtcGxlIHN0YXRlIHZhbHVlLgo=",
      "sub_id": {
          "format": "opaque",
          "id": "f67e39a0a4d34d56b3aa1bc4cff0069f"
       }
    }
  }
}
```
when it should really just be
```
{
  "jti": "123456",
  "iss": "https://transmitter.example.com",
  "aud": "receiver.example.com",
  "iat": 1493856000,
  "sub_id": {
    "format": "opaque",
    "id": "f67e39a0a4d34d56b3aa1bc4cff0069f"
  },
  "events": {
    "https://schemas.openid.net/secevent/ssf/event-type/verification":{
      "state": "VGhpcyBpcyBhbiBleGFtcGxlIHN0YXRlIHZhbHVlLgo="
    }
  }
}
```
like we show in the normalized examples below.

openid-sharedsignals-framework-1_0.md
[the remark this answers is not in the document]
FragLegs: Like my comment above, we should move this out of the event spec and just leave a note about how to specify the `sub_id` at the top level of the SET. This makes it look like `sub_id` should be in the event.

FragLegs (changes requested):

openid-sharedsignals-framework-1_0.md
tulshi: Updated text.

openid-sharedsignals-framework-1_0.md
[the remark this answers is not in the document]
tulshi: Updated text.

openid-sharedsignals-framework-1_0.md
```diff
@@ -240,10 +266,20 @@ Subject Principals are identified by Subject Members defined below.
 
 # Subject Members in SSF Events {#subject-ids}
 
-A member of type Subject in an SSF event MAY have any claim name. Each Subject Member MUST
-refer to exactly one Subject Principal.
+## Subject Members {#subject-members}
+A Subject Member of a SSF event describes a subject of the event. A top-level claim named `sub_id` SHOULD be used to describe the primary Subject Principal of the event.
+
+### Existing CAEP and RISC Events
+Event types already defiend in the CAEP ({{CAEP}}) and RISC ({{RISC}}) drafts MAY use a `subject` field within the `events` claim of the SSF eent to describe the primary Subject Principal of the event. SSF Transmitters SHOULD include the top-level `sub_id` claim even for these existing event types.
```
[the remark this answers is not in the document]
tulshi: fixed.

openid-sharedsignals-framework-1_0.md
tulshi: we have now removed this.

tulshi: This is a breaking change, but since the spec is under development, and this part hasn't been used so far, I thought it was safe to remove the field. We can talk about it on the next call if required.

openid-sharedsignals-framework-1_0.md
tulshi: We can define an interoperability profile which can be more definitive about whether or not this field MUST exist.

timcappalli: I agree that SHOULD needs to be MUST for `sub_id`

appsdesh (commented):
Could you please add following in the section `Distinguishing SETs from other Kinds of JWTs`

> The "sub_id" claim MUST be present, as described in {{{subject-members}}

openid-sharedsignals-framework-1_0.md
tulshi: updated to "MUST"

appsdesh (approved):
This works! Thanks

openid-sharedsignals-framework-1_0.md
```diff
@@ -211,10 +237,20 @@ Subject Principals are identified by Subject Members defined below.
 
 # Subject Members in SSF Events {#subject-ids}
 
-A member of type Subject in an SSF event MAY have any claim name. Each Subject Member MUST
-refer to exactly one Subject Principal.
+## Subject Members {#subject-members}
+A Subject Member of a SSF event describes a subject of the event. A top-level claim named `sub_id` MUST be used to describe the primary Subject Principal of the event.
```
timcappalli, line 241: small nit, but should this be "describes _**the**_ subject of the event" ?

openid-sharedsignals-framework-1_0.md
```diff
@@ -211,10 +237,20 @@ Subject Principals are identified by Subject Members defined below.
...
 
-A Subject may be a "simple subject" or a "complex subject".
+### Existing CAEP and RISC Events
+Event types already defiend in the CAEP ({{CAEP}}) and RISC ({{RISC}}) drafts MAY use a `subject` field within the `events` claim of the SSF event to describe the primary Subject Principal of the event. SSF Transmitters SHOULD include the top-level `sub_id` claim even for these existing event types.
```
timcappalli, line 244: s/defiend/defined

timcappalli: also, should say specifications vs drafts?

openid-sharedsignals-framework-1_0.md
```diff
@@ -1461,6 +1522,37 @@ Cache-Control: no-store
 ~~~
 {: title="Example: Check Stream Status Response" #figstatusresp}
 
+The following is a non-normative example request to check an event stream's
```
FragLegs, line 1525: We intentionally got rid of subject-related status in #88 . Let's not add it back in here.

openid-sharedsignals-framework-1_0.md
```diff
@@ -1527,7 +1619,15 @@ Authorization: Bearer eyJ0b2tlbiI6ImV4YW1wbGUifQo=
 {
   "stream_id": "f67e39a0a4d34d56b3aa1bc4cff0069f",
   "status": "paused",
-  "reason": "Disabled by administrator action"
+  "sub_id": {
```
FragLegs, line 1622: We removed this ability in #88

openid-sharedsignals-framework-1_0.md
```diff
@@ -1541,7 +1641,11 @@ Cache-Control: no-store
 
 {
   "stream_id": "f67e39a0a4d34d56b3aa1bc4cff0069f",
-  "status": "paused"
+  "status": "paused",
+  "sub_id": {
```
FragLegs, line 1645: Removed in #88

openid-sharedsignals-framework-1_0.md
```diff
@@ -1855,31 +1955,60 @@ reason
...
   status.
+  
+As with any SSF Event, this event has a top-level `sub_id` claim:
 
-subject
+sub_id
 
-> REQUIRED. Specifies the stream whose status has been updated.
-  The value of the `subject` field
+> REQUIRED. The top-level `sub_id` claim specifies the Subject Principal for whom the status has been updated.
```
FragLegs, line 1963: In #88 we removed the ability to set status for a specific subject. Therefore the "status updated" event MUST always have a `sub_id` that identifies the stream itself, like the verification event

openid-sharedsignals-framework-1_0.md
```diff
@@ -1855,31 +1955,60 @@ reason
...
+  If the event applies to the entire stream, the value of the `sub_id` field
   MUST be of format `opaque`, and its `id` value MUST be the unique ID of the
   stream.
 
 > Note that the subject that identifies a stream itself is always implicitly
   added to the stream and MAY NOT be removed from the stream.
 
-> Below is a non-normative example of a `stream-updated` event.
+Below is a non-normative example of a `stream-updated` event with a specific subject.
```
FragLegs, line 1971: This first example should be removed. The second one is good though.

openid-sharedsignals-framework-1_0.md
```diff
@@ -217,10 +243,20 @@ Subject Principals are identified by Subject Members defined below.
...
 # Subject Members in SSF Events {#subject-ids}
 
-A member of type Subject in an SSF event MAY have any claim name. Each Subject Member MUST
-refer to exactly one Subject Principal.
+## Subject Members {#subject-members}
+A Subject Member of a SSF event describes a subject of the event. A top-level claim named `sub_id` MUST be used to describe the primary subject of the event.
 
-A Subject may be a "simple subject" or a "complex subject".
+### Existing CAEP and RISC Events
+Event types already defined in the CAEP ({{CAEP}}) and RISC ({{RISC}}) specifications MAY use a `subject` field within the `events` claim of the SSF event to describe the primary Subject Principal of the event. SSF Transmitters SHOULD include the top-level `sub_id` claim even for these existing event types.
```
FragLegs, line 250: Shouldn't this say

> SSF Transmitters MUST include the top-level `sub_id` claim even for these existing event types.

FragLegs (changes requested):

openid-sharedsignals-framework-1_0.md
```diff
@@ -1527,7 +1588,8 @@ Authorization: Bearer eyJ0b2tlbiI6ImV4YW1wbGUifQo=
 {
   "stream_id": "f67e39a0a4d34d56b3aa1bc4cff0069f",
   "status": "paused",
-  "reason": "Disabled by administrator action"
+  "stream_id": "f67e39a0a4d34d56b3aa1bc4cff0069f",
```
FragLegs, line 1591: This line is duplicated

FragLegs (approved):

timcappalli (approved):

Issue #54, opened 2023-04-07
We should include "format" in at least one of the stream configuration examples


FragLegs:
The `format` claim is not present in any of the Stream Configuration request or response examples. It would be helpful to include it at least once so that it is more obvious what that claim is supposed to contain.

PR #87, opened 2023-07-25
Add 'format' to stream config and normative examples in CAEP

FragLegs:
This PR deals with two different uses of `format` 

1. It removes `format` from the stream configuration. Here, `format` means that the stream should expect to only deal with subjects using this format identifier. This closes issue #54
2. It adds `format` to the CAEP event normative examples. Here, `format` is used to denote that the subject in the event is a complex subject. We didn't update this file when we created PR #71


openid-sharedsignals-framework-1_0.md
```diff
@@ -887,7 +887,8 @@ Authorization: Bearer eyJ0b2tlbiI6ImV4YW1wbGUifQo=
     "urn:example:secevent:events:type_2",
     "urn:example:secevent:events:type_3",
     "urn:example:secevent:events:type_4"
-  ]
+  ],
+  "format": "email"
```
appsdesh, line 891: 1. `format` describes how a subject claim should be interpreted. This is already part of the event. This format is abstracted away from the stream configuration. 
2. Stream configuration should define what events it can produce. Adding `format` limits the usage of events in the stream. 

eg. Existing events define `subject` as complex objects like 

```
"subject": {
                "device": {
                    "format": "iss_sub",
                    "iss": "https://idp.example.com/123456789/",
                    "sub": "e9297990-14d2-42ec-a4a9-4036db86509a"
                },
                "tenant": {
                    "format": "opaque",
                    "id": "123456789"
                }
```

In this example an event has 2 subjects of different formats. Adding `format` in the stream will completely break existing event processing as it is more restrictive and not cleanly abstracted.

I do NOT think adding `format` in stream configuration is a clean abstraction

appsdesh (changes requested):

openid-sharedsignals-framework-1_0.md
tulshi: I'm trying to remember when this change was discussed in the WG meetings. Do you have a pointer to the notes of that meeting?

appsdesh: Thanks for clarifying things for me @FragLegs! 

Just like "subjects" are getting removed from the status, I think it would add the value to fix this abstraction/layering issue before the spec is sent for approval. On that note, it may not be needed to update examples at this point.

independentid:
I think there is a general issue of concern that needs more discussion. AFAIK the format is a configuration issue for streams that says what identifier formats and attributes will be used to identify subjects in feeds. SCIM has some variability especially considering ldap and azure to cloud.  I recall google IDP had no clear what to identify subjects as they vary dramatically between RPs.  The the important functionality is that is represents how the issuer and audience have agreed to identify subjects. Yes you could force a single value in the standard but that makes the standard fragile because it becomes limited in usability. For scim I had proposed “Id” and “externalId” is this is what the scim protocol standardizes.  Many have complained that scim is too limited. Others point out that some receivers will not be scim but other systems like ldap or security AI. Subject ID spec and format became the became the way to move forward.  The downside remains how can one build an infinite number of parsers?   The trade off being so you build ssf servers one set of specs and cases or do you add extensibility to handle new format cases over time.  After all this is what consultants and integrators do. PhilOn Jul 28, 2023, at 9:47 AM, Apoorva Deshpande ***@***.***> wrote:﻿
@appsdesh commented on this pull request.



In openid-sharedsignals-framework-1_0.md:
> @@ -887,7 +887,8 @@ Authorization: Bearer eyJ0b2tlbiI6ImV4YW1wbGUifQo=
     "urn:example:secevent:events:type_2",
     "urn:example:secevent:events:type_3",
     "urn:example:secevent:events:type_4"
-  ]
+  ],
+  "format": "email"

Thanks for clarifying things for me @FragLegs!
Just like "subjects" are getting removed from the status, I think it would add the value to fix this abstraction/layering issue before the spec is sent for approval. On that note, it may not be needed to update examples at this point.

timcappalli:
@FragLegs if we're saying that `sub_id` needs to be used, should we refactor all of the examples to use `sub_id`?

tulshi (approved):

appsdesh (approved):

timcappalli (approved):
lgtm

Issue #95, opened 2023-08-03
Authorization server metadata uses `scopes_supported` key vs SSF using `supported_scopes`


appsdesh:
In order to bring consistency and reduce confusion should we align with [RFC8414](https://datatracker.ietf.org/doc/html/rfc8414) to use `supported_scopes` as key in the SSF config metadata?

tulshi:
Actually I thought we deliberately made this different because the intent of those two fields is different. It should be in the meeting notes.

appsdesh:
> Actually I thought we deliberately made this different because the intent of those two fields is different. It should be in the meeting notes.

@atultulshi 

The original [meeting notes from 6/27](https://hackmd.io/@oidf-wg-sse/wg-meeting-20230627) refer the field as `scopes_supported`

In the [following meeting of 7/11](https://hackmd.io/@oidf-wg-sse/wg-meeting-20230711) we restructured the JSON, and made it `supported_scope`. There are no references of consciously changing the naming to avoid conflict, so I assume this is just an oversight.

I agree that it could be confusing, so suggest using completely different naming rather than just flipping 2 words.

Some suggestions -
1. `scopes_required`
2. `scopes_mandatory`

tulshi:
I don't think the intent of the fields is that those scopes are required or mandatory though. They describe scopes that are supported by the AS.

appsdesh:
> I don't think the intent of the fields is that those scopes are required or mandatory though. They describe scopes that are supported by the AS.

The field itself is `OPTIONAL` but if it has scopes populated then it would be mandatory to request the token

appsdesh:
@atultulshi I came across [OAuth 2.0 Protected Resource Metadata](https://datatracker.ietf.org/doc/html/draft-jones-oauth-resource-metadata-04) and it would make more sense to not stuff OAuth properties in the SSF metadata.

This spec suggests having a separate per protected resource metadata for the OAuth server.

mcguinness:
I don't think `OpenID Shared Signals Framework Specification` configuration discovery metadata should directly include OAuth Resource Server metadata like scopes or authorization servers.  There is no requirement that SSF endpoints are OAuth 2.0 Protected Resources.

I see this working similar to SCIM where SSF metadata publishes supported `authenticationSchemes` like https://datatracker.ietf.org/doc/html/rfc7643#section-5

```
"authenticationSchemes": [
   {
     "name": "OAuth Bearer Token",
     "description":
[...]

PR #96, opened 2023-08-03
Use `scopes_supported` key in transmitter metadata to align with RFC8414

appsdesh:
Use `scopes_supported` key in transmitter metadata to align with RFC8414

Resolves issue #95


appsdesh:
@atultulshi @timcappalli @FragLegs I came across [OAuth 2.0 Protected Resource Metadata](https://datatracker.ietf.org/doc/html/draft-jones-oauth-resource-metadata-04) and it would make more sense to not stuff OAuth properties in the SSF metadata.

This spec suggests having a separate per protected resource metadata for the OAuth server.

PR #100, opened 2023-08-09
Do not allow create, update of stream without `events_requested`

### LINKED ISSUE #99: Should we allow `events_requested` to be empty?
`events_requested` is critical for the transmitter to know which events to be transmitted to the receiver. What's the use of the stream if the receiver creates or updates a stream with an empty `events_requested`? 

We should make it clear in the spec and disallow empty `events_requested` in create or update

appsdesh:
Stream does not have any meaning if the `events_requested` is created, updated with an empty array. Fixing the text to advise against such update. 

Resolves #99

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -818,7 +818,7 @@ events_requested
 > **Receiver-Supplied**, An array of URIs identifying the set of events that
   the Receiver requested. A Receiver SHOULD request only the events that it
   understands and it can act on. This is configurable by the Receiver. A
-  Transmitter MUST ignore any array values that it does not understand.
+  Transmitter MUST ignore any array values that it does not understand. This MUST not be an empty list.
 
 events_delivered
 
@@ -878,7 +878,7 @@ the Event Transmitter responds with a "201 Created" response containing a
 The HTTP POST request MAY contain the Receiver-Supplied values of the Stream
 Configuration ({{stream-config}}) object:
 
-* `events_requested`
+* `events_requested` : This MUST not be an empty list
 * `delivery` : Note that in the case of the POLL method, the `endpoint_url` value is
   supplied by the Transmitter.
 * `format`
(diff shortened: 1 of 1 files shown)


FragLegs (changes requested):
Why can't a Receiver set up a stream without events? What if you wanted to create a stream that was dynamically populated with event types at a later point?

FragLegs (changes requested):
Why can't a Receiver set up a stream without events? What if you wanted to create a stream that was dynamically populated with event types at a later point?

appsdesh:
> Why can't a Receiver set up a stream without events? What if you wanted to create a stream that was dynamically populated with event types at a later point?

After the stream is successfully created with relevant events, it makes sense that the requirements change and receiver updates existing `events_requested` with a new array.

What's the use case of stream creation for the empty `events_requested`, what would be the `status` of such a stream.

Maybe we should only prevent updating the stream with empty `events_requested`, if stream creation with empty array is a special case? The receiver can dynamically update the existing stream with the new array of events, there is no need to allow setting it to empty. If the receiver wants to decommission the stream it could do that by the status APIs, rather than making `events_requested` as empty.

Thoughts @FragLegs ?

independentid:
For an update I would assume a missing value means don’t change.A provided value means may be changed. PhilOn Aug 9, 2023, at 10:23 AM, Apoorva Deshpande ***@***.***> wrote:﻿

Why can't a Receiver set up a stream without events? What if you wanted to create a stream that was dynamically populated with event types at a later point?

After the stream is successfully created with relevant events, it makes sense that the requirements change and receiver updates existing events_requested with a new array.
What's the use case of stream creation for the empty events_requested, what would be the status of such a stream.
Maybe we should only prevent updating the stream with empty events_requested, if stream creation with empty array is a special case? The receiver can dynamically update the existing stream with the new array of events, there is no need to allow setting it to empty. If the receiver wants to decommission the stream it could do that by the status APIs, rather than making events_requested as empty.
Thoughts @FragLegs ?

appsdesh:
@independentid in the current shape and form, a replace config call PUT with empty values will reset the config.

FragLegs:
I don't understand what bad thing happens if the Receiver decides to set events_requested to an empty array. True, they won't get any events. But is it our job to police that? They also won't get any events if they never add a subject to the stream. Should we raise alarms in that case too?

I think that the existing behavior is fine as is.

appsdesh:

PR #108, opened 2023-08-30
Stream ID must be unique per stream

appsdesh:
As the transmitter does not have any notion of the receiver uniqueness, it makes more sense to callout the stream ID to be unique per stream.

In the update stream event, `aud` is expected as a stream ID which is expected unique per stream as well.


FragLegs (approved):
This is more clear. Thanks!

tulshi (approved):

openid-sharedsignals-framework-1_0.md
```diff
@@ -788,7 +788,7 @@ following properties:
 stream_id
 
 > **Transmitter-Supplied**, A string that uniquely identifies the stream. Stream
-  IDs MUST be unique per Reciever.  This value is generated by the Transmitter
+  IDs MUST be unique per stream in the context of each transmitter.  This value is generated by the Transmitter
```
tulshi, line 791: "across all streams that are currently open in a Transmitter"

appsdesh: Maybe the following to handle "future" streams, rather than just the "currently open"

> **Transmitter-Supplied**, A string that uniquely identifies the stream. A transmitter MUST generate a unique ID for each of it's streams.

openid-sharedsignals-framework-1_0.md
```diff
@@ -787,9 +787,7 @@ following properties:
 
 stream_id
 
-> **Transmitter-Supplied**, A string that uniquely identifies the stream. Stream
-  IDs MUST be unique per Reciever.  This value is generated by the Transmitter
-  when the stream is created.
+> **Transmitter-Supplied**, A string that uniquely identifies the stream. A transmitter MUST generate a unique ID for each of it's streams at the time of stream creation.
```
tulshi, line 790: it's streams -> its open streams (no apostrophe, and specify open, since the rest of the sentence talks about the time of creation)

appsdesh:

PR #109, opened 2023-08-30
Clarification of events_delivered

appsdesh:
Clarification of events_delivered from the past calls 

The transmitter may reserve rights to deliver events on certain occasions such as (but not limited to) -
1. Receiver has not paid for SKU that the event is part of, 
2. Receiver is not in the desired jurisdiction/compliance environment to receive the event

Discussion contexts -
1. 4/4 https://hackmd.io/@oidf-wg-sse/wg-meeting-20230404 
3. 8/29 https://hackmd.io/nW6HWJ4WQfaX2ZAnm758Uw

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -822,10 +822,12 @@ events_requested
 
 events_delivered
 
-> **Transmitter-Supplied**, An array of URIs which is the intersection of
-  "events_supported" and "events_requested". These events MAY be delivered over
-  the Event Stream. A Receiver MUST rely on the values received in this field
-  to understand which event types it can expect from the Transmitter.
+> **Transmitter-Supplied**, An array of URIs identifying the set of events
+  which is the intersection of "events_supported" and "events_requested". A 
+  transmitter MAY decide to deliver only a subset of events represented by 
+  this intersection. These events MAY be delivered over the Event Stream. A 
+  Receiver MUST rely on the values received in this field to understand which 
+  event types it can expect from the Transmitter.
 
 delivery
 


openid-sharedsignals-framework-1_0.md
```diff
@@ -822,10 +822,12 @@ events_requested
...
-> **Transmitter-Supplied**, An array of URIs which is the intersection of
-  "events_supported" and "events_requested". These events MAY be delivered over
-  the Event Stream. A Receiver MUST rely on the values received in this field
-  to understand which event types it can expect from the Transmitter.
+> **Transmitter-Supplied**, An array of URIs identifying the set of events
+  which is the intersection of "events_supported" and "events_requested". A 
+  transmitter MAY decide to deliver only a subset of events represented by 
+  this intersection. These events MAY be delivered over the Event Stream. A 
+  Receiver MUST rely on the values received in this field to understand which 
+  event types it can expect from the Transmitter.
```
tulshi, line 830: Suggested wording:
events_delivered

**Transmitter-Supplied**: An array of URIs identifying the set of events that the Transmitter SHALL include in the stream. This is a subset of the intersection of "events_supported" and "events_requested". A Receiver MUST rely on the values received in this field to understand which event types it can expect from the Transmitter.

appsdesh: > This is a subset of the intersection of "events_supported" and "events_requested"

I feel the above statement suggests that it will always be a subset, which may not be true. 

If you are referring to "A set is a subset of itself" principle to describe this, I feel it may not be clear to an ordinary reader.

tulshi: If there's a mathematical term to specify the subset that isn't the "proper subset", then we can include that. In normative text, I prefer precision, but happy to consider opinions of other members. Perhaps we can say: "This is a subset (not necessarily a proper subset) of ..."

appsdesh:

PR #117, opened 2023-09-14
Consolidating changes from multiple PRs

appsdesh:
Consolidating changes from PRs

1. https://github.com/openid/sharedsignals/pull/109
2. https://github.com/openid/sharedsignals/pull/108
3. https://github.com/openid/sharedsignals/pull/100


tulshi (approved):

openid-sharedsignals-framework-1_0.md
```diff
@@ -818,13 +818,15 @@ events_requested
 > **Receiver-Supplied**, An array of URIs identifying the set of events that
   the Receiver requested. A Receiver SHOULD request only the events that it
   understands and it can act on. This is configurable by the Receiver. A
-  Transmitter MUST ignore any array values that it does not understand.
+  Transmitter MUST ignore any array values that it does not understand. This
+  array MUST NOT be empty.
```
FragLegs, line 822: Would you consider changing this to SHOULD NOT? Forcing this behavior to raise an error closes off possibilities for the Receiver.

FragLegs: Pulling over this conversation from PR #100 (sorry, I missed that you had responded)

> I don't understand what bad thing happens if the Receiver decides to set events_requested to an empty array. True, they won't get any events. But is it our job to police that? They also won't get any events if they never add a subject to the stream. Should we raise alarms in that case too?
> 
> I think that the existing behavior is fine as is.

> > +1 to keep the spec flexible but we should try to answer following things -
> > 
> > Operationally, a transmitter may decide to cap max number of streams available, a non functional stream like this (one without events subscription) would use up that quota.

Assuming there is a financial (or even just contractual) relationship between the Transmitter and Receiver, a non-functional stream should be accepted and counted towards whatever quota might exist. I am assuming that the no-events state is probably transitory rather than being some final state that requires cleanup of the stream. 

> > If a stream is without any event subscriptions then would it be considered enabled or paused from the status perspective?

I think it would have whatever status is assigned to the stream. Those seem like completely independent values, even if both lead to no events. IT is also the case that not adding any subjects leads to no events. That doesn't change the status of the stream.

> > If we already have a mechanism to pause the stream using update stream then why would we need to support a pseudo pause state by allowing empty subscription?

I don't think setting `events_requested` to an empty array is a pseudo pause. Like you say, we have a pause mechanism. Instead, I'm imagining a situation where the event types a Receiver cares about are dynamically set by some other process. Perhaps they make the selection of event types available in the UI. Or perhaps the event types are added based on other events happening in the system. In both cases, a stream might start without any `events_requested` and have them added over time.

independentid: Regarding events_requested being empty. I suppose events related to the feed itself are still possible. For example the verification event. 

That said, my preference is:
"" or [] means (can't remember if events requested is an array) the client wants no events.  If the attribute isn't present it means no change or "default" (the default set of events preferred by the issuer).

tulshi: @FragLegs I agree we should add language to the spec that the SSF events (stream verification and stream updated) will always be included in the stream, regardless of what is mentioned in the "events_delivered" list. I'm neutral about whether those need to be explicitly included in the "events_delivered" list.

tulshi: @independentid , what does it mean for a client to request that no events be present in the stream? Even if the SSF events (stream verification and stream updated) are permitted, what purpose would such a stream serve?

appsdesh: Updated the text with @FragLegs 's suggestion

independentid: @tulshi honestly I don't really care. I can see however that clients may want to first set up a stream for testing only and then once they think it is configured correctly (e.g. by doing verifications), they can open up the 'floodgates' by modifying events_requested later.

In other words, while not operationally useful, it may be seen as technically useful.  Especially when endpoints are buried inside containers and multiple layers of networks etc.

FragLegs (approved):
Thanks for all the work!

openid-sharedsignals-framework-1_0.md
tulshi: The new changes from Apoorva make this "should not" so this conversation is now outdated. I'm resolving this for now, but please reopen if you think the issue is not yet closed.

PR #123, opened 2023-09-27
incorporated version in the Transmitter Configuration Metadata. Added…

tulshi:
… the Versioning Proposal previously adopted by the WG as a tracked document

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -181,6 +181,15 @@ normative:
     target: https://openid.net/specs/openid-risc-profile-specification-1_0.html
     title: OpenID RISC Profile Specification 1.0 - draft 02
 
+  VERSIONING:
+    author:
+    -
+      ins: A. Tulshibagwale
+      name: Atul Tulshibagwale
+    date: August 2023
+    target: https://openid.github.io/sharedsignals/openid-sharedsignals-versioning-1_0.html
+    title: Shared Signals Versioning Proposal
+
 informative:
   USECASES:
     author:
@@ -563,6 +572,10 @@ configuration information.
 
 Transmitters have metadata describing their configuration:
 
+version
+
+> OPTIONAL. A string representation of a decimal number including a decimal point and at least one number after the decimal point even if it is 0. This indicates the version of the specification the Transmitter adheres to. The version number is determined according to the Versioning ({{VERSIONING}}) specification. The version number of the current specification is "1.0". If the `version` field is missing in the Transmitter Configuration Metadata, then the version number is assumed to be "0.1".
+
 issuer
 
 > REQUIRED. URL using the https scheme with no query or fragment component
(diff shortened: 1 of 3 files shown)


openid-sharedsignals-versioning-1_0.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Protocol Versions
+In the case of the Shared Signals Framework (SSF), a newer draft may change something about the protocol, i.e. the API, event formats, metadata formats, etc. These changes can be categorized as follows:
+
+1. **No Change**: The new version makes no changes to the API, event formats or metadata formats or other properties of the protocol it describes that differs from the Latest Spec.
+2. **Backward Compatible Change**: The new version adds optional properties or methods to the Latest Spec, but an implementation of the Latest Spec that does not implement the new features can continue to operate without seeing any difference in the behavior from the peer that has implemented some or all of the new optional properties.
+3. **Breaking Change**: The new version makes changes to the Latest Spec. Implementations need to be updated in order to be able to work with any implementation of this new version.
+
+**Note** that the change is always compared against the Latest Spec, and not against previous working group drafts.
+Versioning Proposal
```
FragLegs, line 39: Is this a heading? It needs markdown I think

FragLegs (approved):

openid-sharedsignals-framework-1_0.md
```diff
@@ -563,6 +572,10 @@ configuration information.
 
 Transmitters have metadata describing their configuration:
 
+version
+
+> OPTIONAL. A string representation of a decimal number including a decimal point and at least one number after the decimal point even if it is 0. This indicates the version of the specification the Transmitter adheres to. The version number is determined according to the Versioning ({{VERSIONING}}) specification. The version number of the current specification is "1.0". If the `version` field is missing in the Transmitter Configuration Metadata, then the version number is assumed to be "0.1".
```
appsdesh, line 577: - A transmitter may want to support both 0.1 and 1.0 versions. 
- A receiver may want to consume APIs with specific  version

Should the receiver indicate which version it can consume by the headers to the API calls?

appsdesh (changes requested):

openid-sharedsignals-framework-1_0.md
tulshi: I think since the API is incompatible between 0.1 and 1.0, a Transmitter wishing to support both will have to provide two different endpoints, so each metadata can still refer to just one version.

appsdesh: Should the version be part of the API endpoints? Should the API version part of the API URLs or headers? Without that transmitter would not be able to distinguish receivers. 

I feel, there also should be a notion of which version of the spec the receiver supports to allow version negotiation.

tulshi:
Based on WG discussions and discussions with Mike Jones and Mark Haine, this PR is no longer required.

PR #126, opened 2023-10-03
removed duplicate status descriptions

tulshi:


FragLegs (changes requested):
This change removes any description of what the enabled, paused, and disabled states mean. I think we need to at least describe the expected behavior for the Transmitter in each state.

tulshi:
> This change removes any description of what the enabled, paused, and disabled states mean. I think we need to at least describe the expected behavior for the Transmitter in each state.

Do the descriptions in 7.1.2 suffice to describe each state?

FragLegs:
7.1.2 only talks about what the Transmitter should do if it decides to change the stream status without the Receiver's input. It doesn't describe at all what the states mean or what the Transmitter is expected to do in each state.

tulshi:
PR incorrectly removes text that is required in the spec

PR #142, opened 2024-02-27
Clarify verify event response code

appsdesh:
Provide clear guidance on what the response code when the transmitter sends a verify event to the receiver

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -953,7 +953,7 @@ min_verification_interval
 
 description
 
-> **Receiver-Supplied**, An optinal string to describe the properties of the stream.
+> **Receiver-Supplied**, An optional string to describe the properties of the stream.
   This is useful in multi stream systems to identify the stream for human actors. The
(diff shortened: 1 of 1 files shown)


openid-sharedsignals-framework-1_0.md
```diff
@@ -1874,6 +1874,9 @@ publish an SLA for verification event transmission times. Event Receivers MUST N
 depend on the verification event being transmitted synchronously or in any
 particular order relative to the current queue of events.
 
+Upon receiving the verify event, the Reciver should respond with `202 Accepted` to
```
FragLegs, line 1877: This would only make sense if the Receiver had a push-based stream set up. There is no response code for poll-based streams. 

For push-based streams, a 202 is already required by the [push spec](https://datatracker.ietf.org/doc/html/rfc8935#name-success-response), assuming there is nothing wrong with the verify event.

FragLegs (changes requested):

FragLegs (approved):

appsdesh:
Abandoning this

Issue #140, opened 2024-02-09
Allow Receiver to supply public key


FragLegs:
In the current SSF spec, the Transmitter can make a `jwks.json` file available to its Receivers. This allows the Transmitter to sign the JWTs that it is sending the Receiver, so that the Receiver can verify that the security event tokens (SETs) are not forged by some third party. However, these security event tokens often contain personally identifying information (PII) and some vendors may wish to be able to _encrypt_ the SET, not just sign it. In order to do that, the Transmitter would need the Receiver to share a public key. I propose that we add language to the spec to make it clear that a Receiver MAY provide well-known and `jwks.json` endpoints for this purpose.

The current spec treats Receivers as an afterthought - all of the language is around what Transmitters MAY and MUST do. This would give us an opportunity to make Receivers more of a first class citizen in the SSF spec.

appsdesh:
I have raised similar asks around not having reciever's metadata doc in earlier working group calls [here](https://hackmd.io/@oidf-wg-sse/wg-meeting-20230905#SSF-Service-general-reference-instead-of-Transmitter--Receiver)

SET encryption should also be baked in the [SET - RFC8417](https://datatracker.ietf.org/doc/html/rfc8417) or outside SSF spec.

On the encryption - my thoughts are 
1. To encrypt the `events` object, but for that to work, we need to surface stream ID as a first-class claim in the SET JWT that could be used for routing. (Sounds like RFC8417 enhancement)
2. We should define at what granularity we should support encryption, per stream unique keys, per receiver unique keys, etc.

TakahikoKawasaki:
In a traditional way like the [OpenID Connect Dynamic Client Registration 1.0](https://openid.net/specs/openid-connect-registration-1_0.html) specification (OIDC DynReg), an entity that receives encrypted data does either (1) register its public key in advance to the entity that encrypts data (e.g. through the `jwks` client metadata in the case of OIDC DynReg) or (2) make its public key publicly available and convey the location of the key to the entity that encrypts data (e.g. through the `jwks_uri` client metadata in the case of OIDC DynReg).

A new flexible way adopted by the [OpenID for Verifiable Credential Issuance](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html) specification (OID4VCI) is to include a public key for encryption in a request at runtime through a dedicated parameter. In the case of OID4VCI, the parameter is `credential_response_encryption`. The value of the parameter is a JSON object that contains the following properties.

| property | description |
[...]

PR #148, opened 2024-04-02
Optional receiver_key to enable encryption of SETs

FragLegs:
- Added optional `receiver_key` to enable encryption of SETs
- Added language in Security Considerations section to describe when SETs MUST be encrypted
- Added language in Security Considerations offering alternative methods of sharing the receiver's public key, including Dynamic Client Registration

During the WG meeting, we discussed using Dynamic Client Registration out-of-band to share the Receiver's public key with the Transmitter. However, that forces the Transmitter to use OAuth2, which we have intentionally tried to avoid in the spec. So instead I provided an optional in-spec method for sharing the Receiver's public key. For those who would prefer to use Dynamic Client Registration, I also added language to the Security Considerations section describing that as a potential option for secret sharing.

Fixes Issue #140

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -134,6 +134,7 @@ normative:
   RFC7159:
   RFC7517:
   RFC7519:
+  RFC7591:
   RFC8174:
   RFC8414:
   RFC8417:
@@ -574,7 +575,7 @@ spec_version
 
 > OPTIONAL. A version idenitfying the implementer's draft or final specification implemented by the Transmitter. This includes the numerical portion of the spec version as described in the document {{NAMINGCONVENTION}}. If absent, the Transmitter is assumed to conform to "1_0-ID1" version of the specification (this document).
 
->  The following is a non-normative example of Transmitter that implements the second implementer's draft of the Shared Signals Framework specification 1_0. 
+>  The following is a non-normative example of Transmitter that implements the second implementer's draft of the Shared Signals Framework specification 1_0.
 
 ~~~ json
    {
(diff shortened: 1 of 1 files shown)


openid-sharedsignals-framework-1_0.md
```diff
@@ -1005,7 +1013,19 @@ Authorization: Bearer eyJ0b2tlbiI6ImV4YW1wbGUifQo=
     "urn:example:secevent:events:type_3",
     "urn:example:secevent:events:type_4"
   ],
-  "description" : "Stream for Receiver A using events type_2, type_3, type_4"
+  "description": "Stream for Receiver A using events type_2, type_3, type_4",
+  "receiver_key": {
```
FragLegs, line 1017: This example JWK was taken directly from the JWK spec: https://openid.net/specs/draft-jones-json-web-key-03.html#ExampleJWK

openid-sharedsignals-framework-1_0.md
```diff
@@ -957,6 +958,11 @@ description
   This is useful in multi stream systems to identify the stream for human actors. The
   transmitter may truncate the string beyond allowed max length.
 
+receiver_key
```
appsdesh, line 961: 1. We are bringing in receiver-specific wording in the object, which is very stream-specific
2. We should specifically call the key `*_jwks_url` to align with the Tx metadata key and disambiguate its purpose.

appsdesh: For more info, should we also reference the `Encrypting SETs` section?

openid-sharedsignals-framework-1_0.md
```diff
@@ -2036,6 +2128,15 @@ Receivers MUST tolerate receiving events for subjects that have been removed
 from the stream, and MUST NOT report these events as errors to the Event
 Transmitter.
 
+## Encrypting SETs {#management-sec-encrypting-sets}
+If an Event Transmitter is sending events that contain Personally Identifiable Information
+(PII), whether in the body of the event or in the `sub_id` of the SET itself, then the
+Transmitter MUST encrypt the SET such that it is sending a JWE instead of a JWT, as
```
timcappalli, line 2134: While I generally agree that PII should be encrypted, is this really something that the spec should require?

FragLegs: The SET spec already requires it. See the second paragraph here: https://datatracker.ietf.org/doc/html/rfc8417#section-5.1

FragLegs:
Closing and will re-address if needed after release 1.0

PR #152, opened 2024-04-17
Removing transmitter supplied fields from stream config PUT and PATCH…

ysarig1:
Removed the transmitter supplied fields from the stream config PUT and PATCH examples.

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -1282,7 +1282,6 @@ Authorization: Bearer eyJ0b2tlbiI6ImV4YW1wbGUifQo=
 
 {
   "stream_id": "f67e39a0a4d34d56b3aa1bc4cff0069f",
-  "iss": "https://tr.example.com",
   "events_requested": [
     "urn:example:secevent:events:type_2",
     "urn:example:secevent:events:type_3",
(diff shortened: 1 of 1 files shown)


ysarig1:
@FragLegs, @appsdesh, @atultulshi Can you please review the PR and approve/comment?

appsdesh (changes requested):
The existing language already clarifies -

> Transmitter-Supplied properties beside the stream_id MAY be present,
but they MUST match the expected value. Missing Transmitter-Supplied
properties MUST be ignored by the Transmitter.

There may be reasons for the receiver to send the same values to make sure it's info matches the Tx info (and expect to receive error message)

I don't think this PR is needed. Please let me know your thoughts

ysarig1:
> The existing language already clarifies -
> 
> > Transmitter-Supplied properties beside the stream_id MAY be present,
> > but they MUST match the expected value. Missing Transmitter-Supplied
> > properties MUST be ignored by the Transmitter.
> 
> There may be reasons for the receiver to send the same values to make sure it's info matches the Tx info (and expect to receive error message, on mismatch)
> 
> I don't think this PR is needed. Please let me know your thoughts

The issue addressed by this PR is not if the fields can or can't be in the request but the fact that having a the fields in the **examples** is confusing. Having transmitter-supplied fields in an update request is an edge case and while it is allowed, it doesn't make sense to do so.

FragLegs (approved):
Thanks for working on this!

tulshi (approved):

PR #154, opened 2024-04-23
New Session Established Event Type

tulshi:


openid-caep-specification-1_0.md
```diff
@@ -691,6 +706,56 @@ NOTE: The event type URI is wrapped, the backslash is the continuation character
 ~~~
 {: #device-compliance-change-examples-out-of-compliance title="Example: Device No Longer Compliant - Complex Subject + optional claims"}
 
+## Session Established {#session-established}
+Event Type URI:
+
+`https://schemas.openid.net/secevent/caep/event-type/session-established`
+
+The Session Established event signifies that the Transmitter has created a new session for a user for use with the Transmitter. Receivers may use this information for a number of reasons, including:
```
timcappalli, line 714: I think we may need to avoid saying that the "Transmitter has created a new session for a user for use with the Transmitter" because in many cases it will be a different entity that is paired with a transmitter (e.g. Transmitter and IdP are different logical components). We were generic with the text in other events (ex: "Token Claims Change signals that a claim in a token, identified by the subject claim, has changed.")

openid-caep-specification-1_0.md
```diff
@@ -691,6 +706,56 @@ NOTE: The event type URI is wrapped, the backslash is the continuation character
...
+ip
+: The IP address of the user as observed by the Transmitter (**NOTE**, this can be different from the one observed by the Receiver for the same user because of network translation)
+
+fp_ua
+: Fingerprint of the user agent computed by the Transmitter. (**NOTE**, this is not to identify the session, but to present some qualities of the session)
+
+acr
+: The authentication context class reference of the session, as established by the Transmitter. The value of this field MUST be interpreted in the same way as the corresponding field in an OpenID Connect ID Token {{OpenID.Core}}
+
+amr
```
appsdesh, line 732: A session may have been established after multiple factors; a list of AMR claims would help

openid-caep-specification-1_0.md
```diff
@@ -691,6 +706,56 @@ NOTE: The event type URI is wrapped, the backslash is the continuation character
 ~~~
 {: #device-compliance-change-examples-out-of-compliance title="Example: Device No Longer Compliant - Complex Subject + optional claims"}
 
+## Session Established {#session-established}
+Event Type URI:
+
+`https://schemas.openid.net/secevent/caep/event-type/session-established`
```
appsdesh, line 712: I would suggest we add `externalSessionId` as an optional identifier for Tx, Rx to be able to tie in session identities

openid-caep-specification-1_0.md
```diff
@@ -691,6 +706,56 @@ NOTE: The event type URI is wrapped, the backslash is the continuation character
...
+
+fp_ua
+: Fingerprint of the user agent computed by the Transmitter. (**NOTE**, this is not to identify the session, but to present some qualities of the session)
+
+acr
+: The authentication context class reference of the session, as established by the Transmitter. The value of this field MUST be interpreted in the same way as the corresponding field in an OpenID Connect ID Token {{OpenID.Core}}
+
+amr
+: The authentication method reference of the session, as established by the Transmitter. The value of this field MUST be interpreted in the same way as the corresponding field in an OpenID Connect ID Token {{OpenID.Core}}
+
```
appsdesh, line 734: Another imp attribute could be, `Session establishment time` or `createdAt` ( may be different from optional claim `event_timestamp`)

openid-caep-specification-1_0.md
```diff
@@ -691,6 +706,56 @@ NOTE: The event type URI is wrapped, the backslash is the continuation character
...
+### Event Specific Claims {#session-established-event-specific-claims}
+The following optional claims MAY be included in the Session Established event:
+
+ip
+: The IP address of the user as observed by the Transmitter (**NOTE**, this can be different from the one observed by the Receiver for the same user because of network translation)
+
+fp_ua
+: Fingerprint of the user agent computed by the Transmitter. (**NOTE**, this is not to identify the session, but to present some qualities of the session)
+
+acr
```
timcappalli, line 729: Even more important here I think because the values are organization specific (e.g. AAL2 is a NIST term)

openid-caep-specification-1_0.md
```diff
@@ -691,6 +706,56 @@ NOTE: The event type URI is wrapped, the backslash is the continuation character
...
+* An IdP can detect unintended logins
+* A Receiver can establish an inventory of user sessions
+
+### Event Specific Claims {#session-established-event-specific-claims}
+The following optional claims MAY be included in the Session Established event:
+
+ip
+: The IP address of the user as observed by the Transmitter (**NOTE**, this can be different from the one observed by the Receiver for the same user because of network translation)
+
+fp_ua
```
appsdesh, line 726: Do we also need an array to indicate which attributes contributed to the fingerprint? This will help create a contract between Tx and Rx on how to arrive at the fingerprint.

openid-caep-specification-1_0.md
timcappalli: > Another imp attribute could be, `Session establishment time` or `createdAt` ( may be different from optional claim `event_timestamp`)

For `session-revoked`, we said: 

> When event_timestamp is included, its value MUST represent the time at which the session revocation occurred.

We should be consistent.

openid-caep-specification-1_0.md
tulshi: Can this be a part of the metadata? I don't know if we want to include this information in every event.

openid-caep-specification-1_0.md
tulshi: I've not been able to find the right URNs for this. I'm still researching this.

timcappalli: yeah, same. I'm hoping @ve7jtb or @selfissued may be able to help

iamseanodentity: @timcappalli I researched this for some work I am doing and you are spot on with it being org specific as this provides information about the strength of the authentication mechanism or process which pairs with the amr claim nicely. 

Something I was kicking around was urn:<company>:<identity>:loa<1|2|3> (e.g. urn:willywonka:identity:loa1) with a matching amr claim of pwd, as a full example)
I also was trying to incorporate [this](https://openid.net/specs/openid-provider-authentication-policy-extension-1_0.html#auth_policies) but it was a little too rigid.

openid-caep-specification-1_0.md
```diff
@@ -691,6 +708,62 @@ NOTE: The event type URI is wrapped, the backslash is the continuation character
...
+
+
+### Examples {#session-established-examples}
+The following is a non-normative example of the `session-established` event type:
+
+~~~json
+{
+    "iss": "https://idp.example.com/123456789/",
+    "jti": "24c63fb56e5a2d77a6b512616ca9fa24",
+    "iat": 1615305159,
```
iamseanodentity, line 750: @tulshi I think adding in the txn claim supported and defined in RFC 8417 (SET) is warranted here. This alludes to the reconciliation/ledgering aspects of CAEP events. For this CAEP event a user could have multiple sessions established from multiple devices. If there is an unintended session established it could warrant a subsequent `session revoked` and having the txn value match on both SET's is very advantageous from security closure and auditing perspectives.

openid-caep-specification-1_0.md
tulshi: @iamseanodentity  BTW the correct handle on GitHub for me is @tulshi. 

Good point about the Assurance Level Change event. There, we are using a "namespace" field to distinguish between the different assurance level hierarchies. We don't have that here, and we could add some logic about it, or leave it open for now, since there is not registry that we can leverage.

openid-caep-specification-1_0.md
```diff
@@ -691,6 +708,62 @@ NOTE: The event type URI is wrapped, the backslash is the continuation character
...
+    "aud": "https://sp.example.com/caep",
+    "events": {
+        "https://schemas.openid.net/secevent/caep/event-type/session-established": {
+          "subject": {
```
FragLegs, line 754: Can we reformat this example to use the proper `sub_id` in the top level of the SET instead of a `subject` claim in the event?

openid-caep-specification-1_0.md
```diff
@@ -691,6 +708,62 @@ NOTE: The event type URI is wrapped, the backslash is the continuation character
...
+The `event_timestamp` in this event type specifies the time at which the session was established.
+
+### Event Specific Claims {#session-established-event-specific-claims}
+The following optional claims MAY be included in the Session Established event:
+
+ip
+: The array of IP addresses of the user as observed by the Transmitter. The value MUST be in the format of an array of strings, each one of which represents the RFC 4001 {{RFC4001}} string represetation of an IP address. (**NOTE**, this can be different from the one observed by the Receiver for the same user because of network translation)
+
+fp_ua
+: Fingerprint of the user agent computed by the Transmitter. (**NOTE**, this is not to identify the session, but to present some qualities of the session)
```
FragLegs, line 731: For my own understanding, this fingerprint is not meant to be interpretable in any way other than if two fingerprints are equal then the sessions may have come from the same user agent, correct?

openid-caep-specification-1_0.md
FragLegs: Does the Rx need to know anything about how to arrive at the fingerprint? What is the intended meaning of this value (see my comment just below)?

FragLegs (commented):
We don't currently have any way to version the CAEP spec. It is, however, versioned with a 1_0 version tag. Does the addition of new event types require us to update the version? What would updating that version mean?

openid-caep-specification-1_0.md
tulshi: correct.

openid-caep-specification-1_0.md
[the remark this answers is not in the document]
tulshi: I have not found any URNs for this. I think we will have to leave this open for now.

openid-caep-specification-1_0.md
tulshi: Should we add the `txn` claim here, or in SSF? I think the way we're thinking about txn, it will be an SSF-wide change, not specific to the `session-established` event. For the use-case in your comment, should we define a separate claim, e.g., `ext_txn` (for external transaction identifier) and include that in this event?

openid-caep-specification-1_0.md
tulshi: done

openid-caep-specification-1_0.md
FragLegs: I don't think there should be a contract between the Transmitter and Receiver about this value. The Receiver should treat it as an opaque value that can be used to correlate events from the Transmitter, but nothing more.

FragLegs (approved):

openid-caep-specification-1_0.md
tulshi: Agreed.

openid-caep-specification-1_0.md
tulshi: I believe the txn claim should be handled in the SSF spec (thereby including it in all events), so I'm closing this comment here.

tulshi:
> We don't currently have any way to version the CAEP spec. It is, however, versioned with a 1_0 version tag. Does the addition of new event types require us to update the version? What would updating that version mean?

Tim's new proposal would lead us to be able to version each event separately. We can currently refer to the draft numbering for understanding the version number of the CAEP spec.

timcappalli (approved):

Issue #163, opened 2024-05-09
Specify expected default behavior of streams with no subjects


FragLegs:
On a recent WG call, we had a long discussion about the expected behavior of a stream that has not had any subjects added yet. The conclusion was that, by default, a Transmitter should not send any events until a matching subject has been added to the stream. We need to update the spec to indicate this default behavior.

ysarig1:
Unless there is an explicit way to override the default when the stream is created, the "The conclusion was that, by default, a Transmitter should not send any events until a matching subject has been added to the stream" conclusion is very  problematic for enterprise use cases when there may be tens or hundreds of thousands of users and the expectation is that an enterprise admin make the decision for all users.

FragLegs:
Conclusion from 5/14/24 WG call - Add a field in the configuration metadata that indicates whether the Transmitter is an "all subjects by default" or "no subjects by default" transmitter. If not provided, it will default to "no subjects". We will add examples of this in the spec.

PR #168, opened 2024-05-16
Issue 163 - add field to configuration data indicating expected subject behavior for new streams

FragLegs:
Addresses issue #163 by adding an optional field, `default_subjects` to the Transmitter Configuration data. The field is a string which can be either `ALL` or `NONE` and defaults to `NONE` if not provided. The interpretation of this field is as follows:

- `ALL`: All subjects are added by default. Calling `remove_subject_endpoint` acts like a filter, blocking the transmission of only the specified subjects. Calling `add_subject_endpoint` can effectively clear any filters set by `remove_subject_endpoint` but otherwise does nothing.
- `NONE`: No subjects are added by default. Calling `add_subject_endpoint` adds a subject to the stream. Calling `remove_subject_endpoint` can remove a previously added subject, but otherwise does nothing.

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -604,8 +604,9 @@ authorization_schemes
 
 default_subjects
 
-> OPTIONAL. A string indicating the default behavior of newly created streams. The value
-  MUST be either "ALL" or "NONE". If not provided, the default value is "NONE".
+> OPTIONAL. A string indicating the default behavior of newly created streams. If present,
+  the value MUST be either "ALL" or "NONE". If not provided, the Transmitter behavior in
+  this regard is unspecified.
 >  - "ALL" indicates that events for all subjects are transmitted by default. The Receiver
     MAY remove subjects from the stream via the `remove_subject_endpoint`, causing only
     events for those subjects to _not_ be transmitted. The Receiver MAY re-add any
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -593,10 +593,9 @@ default_subjects
   the value MUST be either "ALL" or "NONE". If not provided, the Transmitter behavior in
   this regard is unspecified.
 >  - "ALL" indicates that any subjects that are appropriate for the stream are added to
-    the stream by default. The Receiver
-    MAY remove subjects from the stream via the `remove_subject_endpoint`, causing only
-    events for those subjects to _not_ be transmitted. The Receiver MAY re-add any
-    subjects removed this way via the `add_subject_endpoint`.
+    the stream by default. The Receiver MAY remove subjects from the stream via the
+    `remove_subject_endpoint`, causing events for those subjects to _not_ be transmitted.
+    The Receiver MAY re-add any subjects removed this way via the `add_subject_endpoint`.
 >  - "NONE" indicates that no subjects are added by default. The Receiver MAY add subjects
     to the stream via the `add_subject_endpoint`, causing only events for those subjects
     to be transmitted. The Receiver MAY remove subjects added this way via the
@@ -2173,7 +2172,7 @@ multiple Receivers would lead to unintended data disclosure.
 {: title="Example: SET with array 'aud' claim" #figarrayaud}
 
 ### The "txn" claim {#txn-claim}
-Transmitters SHOULD set the "txn" claim value in Security Event Tokens (SETs). If the value is present, it MUST be unique to the underlying event that caused the Transmitter to generate the Security Event Token (SET). The Transmitter, however, may use the same value in the "txn" claim across different Security Events Tokens (SETs), such as session revoked and credential change, to indicate that the SETs originated from the same underlying cause or reason. 
+Transmitters SHOULD set the "txn" claim value in Security Event Tokens (SETs). If the value is present, it MUST be unique to the underlying event that caused the Transmitter to generate the Security Event Token (SET). The Transmitter, however, may use the same value in the "txn" claim across different Security Events Tokens (SETs), such as session revoked and credential change, to indicate that the SETs originated from the same underlying cause or reason.
 
 ### The "events" claim {#events-claim}
 The "events" claim SHOULD contain only one event. Multiple event type URIs are
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -607,7 +607,8 @@ default_subjects
 > OPTIONAL. A string indicating the default behavior of newly created streams. If present,
   the value MUST be either "ALL" or "NONE". If not provided, the Transmitter behavior in
   this regard is unspecified.
->  - "ALL" indicates that events for all subjects are transmitted by default. The Receiver
+>  - "ALL" indicates that any subjects that are appropriate for the stream are added to
+    the stream by default. The Receiver
     MAY remove subjects from the stream via the `remove_subject_endpoint`, causing only
     events for those subjects to _not_ be transmitted. The Receiver MAY re-add any
     subjects removed this way via the `add_subject_endpoint`.
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -602,6 +602,18 @@ authorization_schemes
   security considerations, make the authorization_schemes attribute
   publicly accessible without prior authentication.
 
+default_subjects
+
+> OPTIONAL. A string indicating the default behavior of newly created streams. The value
+  MUST be either "ALL" or "NONE". If not provided, the default value is "NONE".
+>  - "ALL" indicates that events for all subjects are transmitted by default. The Receiver
+    MAY remove subjects from the stream via the `remove_subject_endpoint`, causing only
+    events for those subjects to _not_ be transmitted. The Receiver MAY re-add any
+    subjects removed this way via the `add_subject_endpoint`.
+>  - "NONE" indicates that no subjects are added by default. The Receiver MAY add subjects
+    to the stream via the `add_subject_endpoint`, causing only events for those subjects
+    to be transmitted. The Receiver MAY remove subjects added this way via the
+    `remove_subject_endpoint`.
 
 TODO: consider adding a IANA Registry for metadata, similar to Section 7.1.1 of
 {{RFC8414}}. This would allow other specs to add to the metadata.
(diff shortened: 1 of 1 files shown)


openid-sharedsignals-framework-1_0.md
```diff
@@ -602,6 +602,18 @@ authorization_schemes
   security considerations, make the authorization_schemes attribute
   publicly accessible without prior authentication.
 
+default_subjects
+
+> OPTIONAL. A string indicating the default behavior of newly created streams. The value
```
tulshi, line 607: If present, the value MUST be either "ALL" or "NONE".

openid-sharedsignals-framework-1_0.md
```diff
@@ -602,6 +602,18 @@ authorization_schemes
...
+  MUST be either "ALL" or "NONE". If not provided, the default value is "NONE".
```
tulshi, line 608: Requiring a certain default behavior when the value is not present will make this backward incompatible. Can we say that if not present, "the Transmitter behavior in this regard is unspecified"

openid-sharedsignals-framework-1_0.md
```diff
@@ -602,6 +602,18 @@ authorization_schemes
...
+>  - "ALL" indicates that events for all subjects are transmitted by default. The Receiver
+    MAY remove subjects from the stream via the `remove_subject_endpoint`, causing only
+    events for those subjects to _not_ be transmitted. The Receiver MAY re-add any
+    subjects removed this way via the `add_subject_endpoint`.
+>  - "NONE" indicates that no subjects are added by default. The Receiver MAY add subjects
+    to the stream via the `add_subject_endpoint`, causing only events for those subjects
+    to be transmitted. The Receiver MAY remove subjects added this way via the
+    `remove_subject_endpoint`.
```
tulshi, line 616: The sentence about adding or removing subjects is common to both initial values in the metadata, and should be specified as one sentence after the descriptions of individual values.

tulshi: understood.

openid-sharedsignals-framework-1_0.md
```diff
@@ -602,6 +602,19 @@ authorization_schemes
   security considerations, make the authorization_schemes attribute
   publicly accessible without prior authentication.
 
+default_subjects
+
+> OPTIONAL. A string indicating the default behavior of newly created streams. If present,
+  the value MUST be either "ALL" or "NONE". If not provided, the Transmitter behavior in
+  this regard is unspecified.
+>  - "ALL" indicates that events for all subjects are transmitted by default. The Receiver
+    MAY remove subjects from the stream via the `remove_subject_endpoint`, causing only
```
jischr, line 611:
```suggestion
    MAY remove subjects from the Stream via the `remove_subject_endpoint`, causing only
```

openid-sharedsignals-framework-1_0.md
```diff
@@ -602,6 +602,19 @@ authorization_schemes
...
+    events for those subjects to _not_ be transmitted. The Receiver MAY re-add any
+    subjects removed this way via the `add_subject_endpoint`.
+>  - "NONE" indicates that no subjects are added by default. The Receiver MAY add subjects
+    to the stream via the `add_subject_endpoint`, causing only events for those subjects
```
jischr, line 615:
```suggestion
    to the Stream via the `add_subject_endpoint`, causing only events for those subjects
```

openid-sharedsignals-framework-1_0.md
```diff
@@ -602,6 +602,19 @@ authorization_schemes
   security considerations, make the authorization_schemes attribute
   publicly accessible without prior authentication.
 
+default_subjects
+
+> OPTIONAL. A string indicating the default behavior of newly created streams. If present,
```
jischr, line 607:
```suggestion
> OPTIONAL. A string indicating the default behavior of newly created Streams. If present,
```

tulshi (approved):

openid-sharedsignals-framework-1_0.md
```diff
@@ -602,6 +602,19 @@ authorization_schemes
   security considerations, make the authorization_schemes attribute
   publicly accessible without prior authentication.
 
+default_subjects
```
appsdesh, line 605: @FragLegs @atultulshi Though this will help keep things backward compatible, imo this introduces stream specific behavior being defined in 2 separate places -
1. Config Metadata - Whether all subjects are included in the bootstrapped stream
2. Stream create/update requests - Which events should be part of the stream

Thoughts on [this](https://github.com/openid/sharedsignals/issues/163#issuecomment-2110197678) suggestion?

appsdesh: > Do you really think the way a Transmitter treats subjects is stream-specific?
In the case of an IdP (tx) - Service Provider (Rx) relationship, an SP may need to consume events only assigned users. Eg. In case of Duo IdP (Tx) -> Splunk (Rx), Splunk would not want to hear events from non-engineers who never access Splunk beginning from the stream inception.

> It seems strange to modify the stream itself, which should be more constant, in those situations.
My suggestion was only for the stream bootstrap scenario. For future changes, the receiver can continue using add/remove subjects 

> We don't want to return the subjects in the StreamConfiguration data. That would make the PUT and PATCH calls quite difficult if you need to update the stream for any reason.
I agree; we can skip adding subjects to the response. Only keep it in the request to indicate what the stream should bootstrap with. In any case, we do not provide a way for the receiver to know which subjects are part of the stream, Rx is expected to maintain such a state.

> I'm not sure how we would offer the equivalent of the default ALL under your proposal.
All may be shown by the subject as an entire tenant.

FragLegs: > In the case of an IdP (tx) - Service Provider (Rx) relationship, an SP may need to consume events only assigned users. Eg. In case of Duo IdP (Tx) -> Splunk (Rx), Splunk would not want to hear events from non-engineers who never access Splunk beginning from the stream inception.

This is a good example for why a Tx might want to default to NONE and let the Rx add specific subjects. What I'm having trouble imagining is that Duo might want to default to NONE for Splunk but then default to ALL for Okta. It seems like an engineering architecture decision (i.e. how you build your transmitter) rather than a stream-by-stream decision.

> My suggestion was only for the stream bootstrap scenario. For future changes, the receiver can continue using add/remove subjects

Having two ways to add subjects seems like it would add a lot of confusion.

> All may be shown by the subject as an entire tenant.

That would still need to be a thing that the Receiver requests in your scenario. It's just that they are requesting it during stream configuration instead of during an add_subject call. What this PR is adding is a way for the Transmitter to say, "By default, we will send all subjects and the Receiver does not have to do anything special." This is the behavior that multiple members of the working group asked for (including you, I thought) during our call.

I'll also note that the interop spec says that "Receivers MUST assume that all subjects are implicitly included in a Stream, without any `AddSubject` method invocations." If Transmitters are going to act that way, I think it would be nice for them to be able to indicate it (although nothing in this PR forces them to indicate it - the new field is entirely optional). I know that other Transmitters do not want to include all subjects implicitly. My interest here is in making that difference visible to a Receiver.

tulshi: I missed this discussion earlier, sorry. @appsdesh , your [suggestion](https://github.com/openid/sharedsignals/issues/163#issuecomment-2110197678) is a good one, but I don't think it is in conflict with this PR. For streams that define their default behavior to be NONE, they could still use the AddSubject in the way you suggest (or add a parameter to the CreateStream method) to specify the locus of subjects they are interested in receiving events about. However I think this is a bigger change (the re-interpretation of AddSubject and the ability to potentially define a subject pattern in CreateStream), so we should take it up after v1.0. Thoughts. Also, I do not think that this PR creates the situation where stream related information is in two places. The Transmitter Configuration Metadata already includes things like "delivery_methods_supported", which are stream related.

timcappalli:
Deferring to @appsdesh on this one.

openid-sharedsignals-framework-1_0.md
appsdesh: @FragLegs 

> Duo might want to default to NONE for Splunk but then default to ALL for Okta.

As I understand, SCIM faces a similar issue and solves it by having  _the Transmitter decide_ which subjects are valid on the stream/connection. Our specification already suggests that Rx should reject anything that it does not understand.

@tulshi 

`delivery_methods_supported` is a handshake requirement between a Tx and Rx, we usually won't anticipate the same pair of Tx-Rx doing 2 streams, one on PUSH and other on PULL. So I would not call it a stream property. 

I am ok with not doing _any_ changes related to this and bringing it up post-v1

tulshi:
> Deferring to @appsdesh on this one.

@timcappalli You would need approve in order for us to be able to merge (or @iamseanodentity can)

openid-sharedsignals-framework-1_0.md
FragLegs: > `delivery_methods_supported` is a handshake requirement between a Tx and Rx, we usually won't anticipate the same pair of Tx-Rx doing 2 streams, one on PUSH and other on PULL. So I would not call it a stream property.

The exact same argument could be made about the proposed `default_subjects` field. It is a handshake requirement in that it defines how the Receiver should expect the Transmitter to perform. It would be _baffling_ for it to be different for 2 streams in the same Tx-Rx pair.

I honestly don't understand your objection to this PR. It is not proposing a new way of managing subjects. It is just trying to make explicit something that already exists but is currently left implicit and undefined.

Perhaps it is confusion about the name of the field. Would it make more sense if the field was called `default_subject_behavior`?

appsdesh: > It is a handshake requirement in that it defines how the Receiver should expect the Transmitter to perform. It would be baffling for it to be different for 2 streams in the same Tx-Rx pair.

This is _NOT_ a handshake property, i.e., the absence of this property would not fail the stream establishment, and events would flow. 

A Tx-Rx pair legitimately may have reasons to define one stream with all subjects and another with none, as described above. 

The point I am trying to make is - this is not something that would 
1. Fail the bootstrapping process if it's not available; hence may not be a constituent of discovery/Config Metadata doc.
2. Tx may need to bootstrap streams with various subjects, ALL, NONE, and Subset, based on who the receiver is. Just two values may prevent future flexibility on how we bootstrap streams.

iamseanodentity:
I just caught up on the fireworks :)

tulshi (approved):

appsdesh (approved):
Thanks for clarifying it

Issue #141, opened 2024-02-09
Mark fields optional or required in StreamConfiguration


FragLegs:
In the StreamConfiguration section, the `description` field says:

> An optinal string to describe the properties of the stream. This is useful in multi stream systems to identify the stream for human actors. The transmitter may truncate the string beyond allowed max length.

This is the only StreamConfiguration field that says anything about whether the field is optional or not. We should either:
1. Infer from this that all of the other fields must be required
2. Be explicit about which fields are required/optional

PR #170, opened 2024-05-16
Add OPTIONAL/REQUIRED to the fields in the stream configuration

FragLegs:
Addresses Issue #141 . When we added `description` to the Stream Configuration we said that it was an optional field. That left open a question about whether the other fields were required or not. The descriptions on the fields make it clear that some of them should be optional. This PR makes it explicit which ones are optional and which are required.


timcappalli (approved):
lgtm

tulshi (approved):

Issue #162, opened 2024-05-07
Issuer Mix-Up


PedramHD:
This issue is part of the ongoing formal security analysis of the SSF specification. We already created a formal model of the specification, identified and formalized security properties, and described the assumptions (as also agreed upon with the Working Group) upon which the analysis is based here: https://github.com/openid/sharedsignals/files/15014966/2024-04-09_WP4.1a-Report.pdf

### Setting
As detailed in our report, we assume that Transmitters identify Receivers by dynamically created nonces which the Transmitter issues when responding to a configuration discovery request. More precisely, we assume that the Transmitters provide different endpoints for each configuration discovery request, with the URLs containing a unique part `u`, which is also used as the audience value for this Receiver. However, this issue also applies more generally to cases in which the identifier is chosen by the Transmitter for the Receiver out-of-band and transmitted from the Receiver to the Transmitter in a way that both parties agreed upon.

### Overview
Let `t` be an honest Transmitter and `r` an honest Receiver, with `u_r` being the audience value assigned from `t` to `r`. Furthermore, let `att` be the attacker, and let `u_att` be the audience value that `t` assigned to `att`.
Intuitively, `r` should not accept a SET created by `t` with the audience value of the attacker. As shown in the following, this might happen under the setting described above, breaking the Session Integrity property from our report (see Section 4.2).

### Detailed Description

We now give more details, based on the following diagram:
 
![issuer-mix-up](https://github.com/openid/sharedsignals/assets/45629876/74c99661-7fe5-453d-937b-4b218a639f60)

- Configuration Discovery
  - Both the honest Receiver `r` and the attacker `att` get the configuration discovery metadata of `t` (Steps 1 - 4 of the diagram). As noted previously, we assume that the Transmitter assigns the audience values `u_r` and `u_att`.
  - The Receiver `r` obtains the configuration from `att` in Steps 5-6. Here, the attacker re-uses the audience value that it obtained from `t`, i.e., `u_att`.
  - The Receivers obtain the JWKS from the corresponding endpoints (not shown in the diagram).
- Stream Creation using Poll Delivery
  - The Receiver `r` sends a request to `att` to create a new stream (see Step 7)
  - The attacker creates a stream at `t` and receives the polling endpoint `pollEP_t` and the stream identifier `s_att` (see Steps 8 - 9).
  - The attacker responds to the stream creation request with the same stream configuration that it obtained from `t` in Step 10.
[...]

PR #172, opened 2024-05-23
Add SHOULD language about checking the issuer value

FragLegs:
To address the attack identified in Issue #162, we recommend that the Receiver SHOULD check the `iss` value when getting a Stream Configuration from the Transmitter to ensure that it matches the Issuer that sent the Transmitter Configuration data.

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -927,7 +927,9 @@ In order to communicate events from a Transmitter to a Receiver, a Receiver
 MUST first create an Event Stream. An Event Receiver creates a stream by making
 an HTTP POST request to the Configuration Endpoint. On receiving a valid request
 the Event Transmitter responds with a "201 Created" response containing a
-[JSON][RFC7159] representation of the stream’s configuration in the body.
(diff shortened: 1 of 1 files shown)


appsdesh (commented):
Along with this, the receivers MUST validate `iss` claims on every SSF event delivered on the stream

FragLegs:
> Along with this, the receivers MUST validate `iss` claims on every SSF event delivered on the stream

@appsdesh I agree that Receivers ought to validate the `iss` claim in the SET, but is there a specific attack that you are imagining that can only be countered this way?

tulshi (approved):

timcappalli:
Why is this a SHOULD vs a MUST?

FragLegs:
> Why is this a SHOULD vs a MUST?

@timcappalli I don't feel strongly about this, but the SSF spec is mostly concerned with what the Transmitter MUST do. It seems like we generally take a more lenient stance towards the Receiver. I think the underlying statement is, "If the Receiver doesn't want to be subjected to attacks a, b, and c, then it SHOULD do x, y, and z."

FragLegs:
Note to self: also need to tie Issuer of metadata to iss

timcappalli (approved):

Issue #161, opened 2024-05-07
Stream Audience Mix-Up


PedramHD:
The SSF specification does not require authentication of Receivers at Transmitters. Under certain conditions, this enables an attacker to get a SET for audience values of other Receivers.

Please note that this issue is part of the ongoing formal security analysis of the SSF specification. We already created a formal model of the specification, identified and formalized security properties, and described the assumptions (as also agreed upon with the Working Group) upon which the analysis is based here: https://github.com/openid/sharedsignals/files/15014966/2024-04-09_WP4.1a-Report.pdf

### Setting

In the following, we briefly summarize the relevant assumptions from our report.

1. There is no Receiver authentication, as the SSF specification has no requirements on authentication of the Receiver at the Transmitter.
2. The audience values of the Receivers are not secret and, therefore, known to the attacker. (For example, the audience value can be a domain of the Receiver, a number that the Transmitter increases for each new Receiver, or some low-entropy value that is easily guessable.)
3. Transmitters can distinguish between different Receivers when receiving requests to the management endpoint, e.g., by providing different URLs to each Receiver (these URLs essentially contain the audience value).
4. Subjects are added to a stream based on authorization tokens contained in the request to the add subject endpoint.

### Detailed Description
Under this setting, the attacker can trivially get SETs for the audience value of an honest Receiver: Let `u` be the identifier (i.e., the audience value) of an honest Receiver `r` at an honest Transmitter `t`. Without authentication, the attacker can create a new stream at `t` for the audience value `u`, and can therefore (at a later step) receive a SET with this audience. (The attacker would potentially need to first add subjects to the stream, which can happen for subjects for which the attacker is authorized). This breaks the confidentiality property described in our report (see Section 4.3). This would be particularly problematic if, upon creating a stream for `u`, the Transmitter would add certain subjects that only `r` is authorized to access information about.

### Possible Fix
One possible fix is to mandate authentication at the management API, i.e., the Transmitter would use the authenticated identity value as the audience of the SET.

Issue #160, opened 2024-05-07
Attacker Stream Subject Insertion


PedramHD:
Under certain conditions, an attacker might receive SETs for subjects without being authorized to access information on these subjects. We give the details in the following.

Please note that this issue is part of the ongoing formal security analysis of the SSF specification. We already created a formal model of the specification, identified and formalized security properties, and described the assumptions (as also agreed upon with the Working Group) upon which the analysis is based here: https://github.com/openid/sharedsignals/files/15014966/2024-04-09_WP4.1a-Report.pdf

### Setting

We start with a brief summary of the relevant assumptions from our report.

1. There is no Receiver authentication, as the SSF specification has no requirements on authentication of the Receiver at the Transmitter.
2. Transmitters identify Receivers by dynamically created nonces which the Transmitter issues during configuration discovery. More precisely, we assume that the Transmitter provides different endpoints for each configuration discovery request, with the URLs containing a unique part `u`, which is also used as the audience value for this Receiver. However, this issue also applies more generally to cases in which the identifier is chosen by the Transmitter for the Receiver out-of-band and transmitted from the Receiver to the Transmitter in a way that the parties agreed upon.
3. Subjects are added to a stream based on authorization tokens contained in the request to the add subject endpoint. We assume that these tokens are pre-shared and that receivers select the appropriate token to include in a request based on the service, i.e., the endpoint to which the request is sent.

### Detailed Description
The attacker receives a SET with information on a subject that the attacker is not authorized to receive (i.e., breaking the authorization property described in Section 4.4 of our report). This happens by a mix-up in which the attacker acts as a Receiver towards an honest Transmitter `t` and as a Transmitter towards an honest Receiver `r`.
At first glance, it might seem unreasonable to expect security guarantees if a Receiver is interacting with a malicious Transmitter. However, even in this case, the malicious Transmitter should not be able to compromise the security guarantees between the Receiver and any other Transmitter that the Receiver is using.
Let `iss_t` be an issuer identifier of the Transmitter `t` and `iss_att` an issuer identifier of the attacker. Furthermore, let `at_subID` be an access token that the Receiver `r` shares with the Transmitter `t`, and which authorizes access to the subject `subID`, i.e., upon receiving this access token at the add subject endpoint, the Transmitter would add `subID` to the corresponding stream. Furthermore, we assume that only `r` is authorized to access information about `subID`.
[...]

PR #134, opened 2023-12-21
Include OAuth specifics in the interop spec

appsdesh:
We have defined the use of OAuth 2.0 as an authorization scheme in our interoperability draft. 

This PR adds more details about using OAuth constructs to make vendor interoperability more standardized.

PR #173, opened 2024-05-23
Added language requiring authorization of stream management API

FragLegs:
To address the attacks proposed in #161 and #160, this PR adda a paragraph indicating that all Stream Management API endpoints must use authorization that associates stream IDs with a specific Receiver, unless some other method of trust is established.

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -96,9 +96,11 @@ normative:
     title: OpenID Connect Core 1.0 - ID Token
   OASIS.saml-core-2.0-os:
   RFC2119:
+  RFC2818:
   RFC6749:
   RFC6750:
   RFC7159:
+  RFC7235:
   RFC7517:
   RFC7519:
   RFC8174:
@@ -761,6 +763,14 @@ This section defines an HTTP API to be implemented by Event Transmitters
 which can be used by Event Receivers to create and delete one or more Event Streams.
 The API can also be used to query and update the Event Stream's configuration and status,
 add and remove Subjects, and trigger verification for those streams.
+
+Unless there exists some other method of establishing trust between a Transmitter and
+Receiver, all Stream Management API endpoints MUST use HTTP over TLS {{RFC2818}}
+and standard HTTP authentication and authorization schemes, as per {{RFC7235}}.
+The authorization MUST scope a Receiver to specific stream IDs, such that only that
+Receiver is able to access or modify the details of the Event Stream.
+The Transmitter MAY describe the available authorization schemes in the Transmitter
+Configuration metadata, as described in {{discovery-meta}}.
 
 ~~~
 +------------+                +------------+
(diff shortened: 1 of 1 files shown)


openid-sharedsignals-framework-1_0.md
```diff
@@ -762,6 +764,14 @@ which can be used by Event Receivers to create and delete one or more Event Stre
 The API can also be used to query and update the Event Stream's configuration and status,
 add and remove Subjects, and trigger verification for those streams.
 
+Unless there exists some other method of establishing trust between a Transmitter and
```
appsdesh, line 767: https://github.com/openid/sharedsignals/pull/134/files adds TLS language to interop spec, but 👍 to bring it here.

Though, I feel the last sensence around `authorization schemes` is either duplicate of what is already referenced above or misplaced.

openid-sharedsignals-framework-1_0.md
```diff
@@ -762,6 +764,12 @@ which can be used by Event Receivers to create and delete one or more Event Stre
...
+Receiver, all Stream Management API endpoints MUST use HTTP over TLS {{RFC2818}}
+and standard HTTP authentication and authorization schemes, as per {{RFC7235}}.
+The authorization MUST scope a Receiver to specific stream IDs, such that only that
```
tulshi, line 770: I'm not sure about a few things here:

- Using "scope" as a verb. Should we say "MUST restrict"? That raises the question how should it restrict. I'd suggest using language like "The authorization scheme used MUST be capable of restricting a stream to only one Receiver"
- The bigger point I'm unsure about is whether we should have the "MUST" language here at all. It is possible for a situation like: Receiver A has a stream for all "token-claims-change" events. Receiver B has a stream for "assurance level change" events. But both Receiver A and Receiver B share the stream for "session-revoked" event. They can both effect the required changes on receiving the event, so it doesn't matter to them which Receiver receives it.

FragLegs: Both points make sense. I'll update the "scope" language and broaden the protection so that the Transmitter can use the auth to associate a Receiver with one or more streams without requiring the association be 1:1.

openid-sharedsignals-framework-1_0.md
```diff
@@ -762,6 +764,12 @@ which can be used by Event Receivers to create and delete one or more Event Stre
 The API can also be used to query and update the Event Stream's configuration and status,
 add and remove Subjects, and trigger verification for those streams.
 
+Unless there exists some other method of establishing trust between a Transmitter and
+Receiver, all Stream Management API endpoints MUST use HTTP over TLS {{RFC2818}}
+and standard HTTP authentication and authorization schemes, as per {{RFC7235}}.
+The authorization MUST associate a Receiver with one or more stream IDs, such that only
```
timcappalli, line 770: "The authorization" reads a bit weird to me but not sure I have a better suggestion.

timcappalli (approved):

openid-sharedsignals-framework-1_0.md
FragLegs: Would you prefer "This authorization"?

FragLegs:
Update RFC2818 and RFC7235 to RFC9110

timcappalli (approved):

tulshi (approved):

Issue #166, opened 2024-05-14
Initial security audit feedback


FragLegs:
The first draft of the security audit highlighted several small items that ought to be fixed before we move to a release version. 

[2024-04-09_WP4.1a-Report.pdf](https://github.com/openid/sharedsignals/files/15309694/2024-04-09_WP4.1a-Report.pdf)

The relevant sections have been copied below.

# 3. Notes on the SSF Specification
Based on our work with the SSF specification, we suggest the working group consider the following
changes. Most of these are based on (necessary) assumptions explained in Section 2 and we note
that our model reflects the specification with these changes.
## Issuer Identifier Validation. 
While the configuration discovery specification mandates Transmitters to only use issuer identifiers with the https scheme [10, Section 6.1], there is no requirement
for Receivers to only request configuration documents from https URLs. If a Receiver requests a
Transmitter’s configuration document from an http URL, a network attacker may launch an MitM
attack, resulting in the Receiver accepting arbitrary, attacker-chosen configuration data (including
JWKs) and arbitrary, attacker-chosen SETs. We, therefore, recommend explicitly mandating Receivers to (1) obtain Transmitters’ issuer identifiers from trusted sources, and (2) verify that these
issuer identifiers use the https scheme.
In our model, we assume that these recommendations are implemented (see Section 2.5).
## Stream Configuration Management API Endpoints. 
The current SSF specification does not
mandate the use of https for any of the stream configuration management API endpoints. We
note that for SET delivery, [RFC8935, RFC8936] mandate the use of https URLs, and of course
recommend mandating Transmitters to use https URLs for all stream management API endpoints
and the JWKs endpoint.

In our model, we assume that this recommendation is implemented (see Section 2.5).
## Requirements on Additional SET Delivery Methods. 
In our model, we only consider the push
and poll delivery methods as defined in [RFC8935, RFC8936] (see Section 2.2). This implies that
in our model, SET delivery always takes place via TLS-protected connections. However, the SSF
5
specification does not limit allowed delivery methods to these two. Hence, we recommend requiring
the use of SET delivery methods that ensure SET confidentiality and integrity.

FragLegs:
> Issuer Identifier Validation

I will address this in an associated PR.

> Stream Configuration Management API Endpoints

PR #173 requires TLS for all stream management endpoints

> Requirements on Additional SET Delivery Methods

[...]

PR #174, opened 2024-05-24
Add sentence suggesting that Issuer information should be validated by the Receiver

FragLegs:
There are three problems mentioned in Issue #166:

1) We need to add language that suggests the Receiver validate the Issuer value before doing discovery. This PR addresses that.

2) We need to add TLS restriction to the stream management endpoints. PR #173 addresses that.

3) We need to restrict delivery methods to only secure options. That is already done by the fact that the Stream Configuration metadata's `delivery.method` field can only be one of `push` or `poll`.

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -632,7 +632,7 @@ Transmitter.
 
 Using the Issuer URL as documented by the Transmitter, the Transmitter Configuration
 Metadata can be retrieved. Receivers SHOULD ensure that the Issuer URL comes from a
-trusted source and uses the https scheme.
+trusted source and uses the `https` scheme.
 
 Transmitters supporting Discovery MUST make a JSON document available at the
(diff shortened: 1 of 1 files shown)


tulshi (approved):

openid-sharedsignals-framework-1_0.md
```diff
@@ -631,7 +631,8 @@ Transmitter.
 ## Obtaining Transmitter Configuration Metadata
 
 Using the Issuer URL as documented by the Transmitter, the Transmitter Configuration
-Metadata can be retrieved.
+Metadata can be retrieved. Receivers SHOULD ensure that the Issuer URL comes from a
+trusted source and uses the https scheme.
```
timcappalli, line 635:
```suggestion
trusted source and uses the `https` scheme.
```

timcappalli (changes requested):

tulshi (approved):

timcappalli (approved):

Issue #135, opened 2023-12-26
Bring setting and reading stream status into alignment


FragLegs:
When you read a stream's status (GET), you get back a response with two fields:

- status
- reason

When you set a stream's status (POST), you send an object with three fields and get the same object back:

- stream_id
- status
- reason

Transmitter code would be cleaner if the GET endpoint also returned a `stream_id` in the response object.

PR #175, opened 2024-05-24
Add stream_id to the response when getting stream status

FragLegs:
This PR addressed issue #135 by adding `stream_id` to the response of a GET request for stream status. The response for a POST request already includes `stream_id` and it makes implementations easier if the GET and POST requests return the same object.

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -1466,6 +1466,10 @@ stream_id
 On receiving a valid request, the Event Transmitter responds with a 200 OK
 response containing a [JSON][RFC7159] object with the following attributes:
 
+stream_id
+
+> REQUIRED. The stream whose status is being queried.
+
 status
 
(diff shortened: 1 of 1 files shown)


tulshi (approved):

openid-sharedsignals-framework-1_0.md
```diff
@@ -1466,6 +1466,10 @@ stream_id
 On receiving a valid request, the Event Transmitter responds with a 200 OK
 response containing a [JSON][RFC7159] object with the following attributes:
 
+stream_id
+
+> REQUIRED. The stream whose status is being queried.
+
```
appsdesh, line 1472: nit: to add more clarity should we say

> The identifier of the stream whose status is being queried

and change it throughout the spec

FragLegs: Good catch

appsdesh (approved):

Issue #167, opened 2024-05-14
Clarify that only one event is allowed per SET


FragLegs:
During the 5/14/24 WG meeting we discussed the need to clarify that only one event (key/value pair in the `events` claim) is allowed per SET.

timcappalli:
Since we're now requiring the subject to be at the top level of the SET, should we consider allowing multiple events around the same subject? The request has come up in the past.

jischr:
i see a lot of advantages to allowing multiple events in larger scale implementations, running bulk jobs, etc.

@atultulshi what are your thoughts on this?

tulshi:
I agree there are benefits to be able to send multiple events. Whether we want to do this before 1.0 is a question I'm trying to grapple with. My (weakly held) opinion is that we should move this to be after the 1.0 spec.

independentid:
The stipulation in SET rfc is multiple events are possible as long as it is just different statements about the same transaction.  Eg you could gave both a scim state change event and a risc signals event from the same transaction. The trade off of splitting into single event payloads, receivers may assume two or more events are multiple transactions unless they check the txn value. Multiple different transactions in a set are not allowed.  Multiple payloads are. PhilOn May 16, 2024, at 8:11 AM, Jen Schreiber ***@***.***> wrote:﻿
i see a lot of advantages to allowing multiple events in larger scale implementations, running bulk jobs, etc.
@atultulshi what are your thoughts on this?

FragLegs:
Closed by https://github.com/openid/sharedsignals/pull/179

PR #179, opened 2024-06-04
Specify one event per "events" claim of SET for Interoperability Profile

jischr:
As discussed in WG meeting from 2024-05-14, specify that only one event is allowed in the interoperability profile.

Fixes Issue #167

### DIFF
--- a/openid-caep-interoperability-profile-1_0.md
+++ b/openid-caep-interoperability-profile-1_0.md
@@ -179,8 +179,7 @@ All events MUST be signed using the `RS256` algorithm using a minimum of 2048-bi
 ## Security Event Token
 
 ### The "events" claim
-The "events" claim of the SET MUST contain only one event. If more than one event is provided in the "events" claim, only the first event will be respected. All other events
-will be ignored.
+The "events" claim of the SET MUST contain only one event.
 
 # Use Cases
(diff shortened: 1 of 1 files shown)


openid-caep-interoperability-profile-1_0.md
```diff
@@ -176,6 +176,11 @@ Receivers MUST be prepared to accept events with any of the subject identifier f
 ## Event Signatures
 All events MUST be signed using the `RS256` algorithm using a minimum of 2048-bit keys.
 
+## Security Event Token
+
+### The "events" claim
+The "events" claim of the SET MUST contain only one event.
+
```
jischr, line 183: I was a bit back and forth on adding a follow up:
```
If more than one event is provided in the "events" claim, only the first element in the "events" will be respected. All other events will be ignored.
```

This seemed to get more confusing as a JSON object is an unordered collection so the "first" event is a meaningless concept. I left it out but happy to add some variation of it back.

FragLegs (approved):

timcappalli (approved):

tulshi (approved):

openid-caep-interoperability-profile-1_0.md
tulshi: I'm OK with the simpler language that Jen has proposed. Not sure what value we get by adding this language about additional events.

Issue #157, opened 2024-04-24
Update CAEP & RISC Events and SSF Docs with txn claim


iamseanodentity:
The `txn` claim is called out as optional on RFC 8417 but is not referenced on CAEP / RISC Events or in the general SSF documentation. With the new CAEP Event being introduced, `session established`, it feels like introducing `txn` now is the right time before v1 is established. This speaks to how a transmitter(SST) and receiver(SSR) can co-op using a standard JWT claim. 

SST -> session revoked CAEP Event -> SSR with txn: 123. The SSR that received the SET can then send back, acting as a SST, a session revoked event to the Transmitter which is now acting as a SSR. This is a good example of auditing and accounting practices. This also helps inform a SST that the signals it is emititng to the SSR's are still accurate and not subjective noise to a SSR (i.e. the underlying data powering the SST is valid and accurate...which is important). 

Would like to update the open-id-caep-spec (Sections 2 and 3), open-id-risc-spec (Sections 2 and 3) and open-id-sharedsignals-framework-1.0 (Section 5) md files. In the events I think we would want to call out the specifics to use the `txn` claim and not have it optional for _all_ events.

FragLegs:
For clarity's sake, the proposal here is to:
1. Update the SSF spec to make `txn` a _required_ top-level claim of the SET for all SSF events (top-level meaning it nests at the same level as `iss` and `aud` in the SET json).
2. Update the SSF spec to include `txn` in any non-normative examples of events.
3. Update the CAEP and RISC specs to include `txn` in the non-normative examples.

Is that correct?

iamseanodentity:
PR submitted

FragLegs:
Closed by https://github.com/openid/sharedsignals/pull/180

PR #180, opened 2024-06-06
PR for Issue #157 (Update CAEP & RISC Events and SSF Docs with txn claim)

iamseanodentity:
Added txn claims to examples and made a direct callout to the SET RFC and called out the explicit use of the claim.

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -2017,9 +2017,6 @@ The CAEP use cases that set the requirements are described in CAEP Use Cases (TO
 This section provides SSF profiling specifications for the Security Event Token (SET)
 {{RFC8417}} spec.
 
-{{RFC8417}} includes the usage of the `txn` claim in the specification. Transmitters SHOULD set the `txn` value in the Security Event Tokens (SETs). If the value is present, it MUST be unique to the underlying event that caused the Trasmitter to generate the Security Event Token (SET). The Transmitter, however, may use the same `txn` value across different Security Events Tokens(SETs), such as session revoked and credential change, to indicate that they originated from the same underlying cause or reason. 
-
-
 ### Signature Key Resolution {#signature-key-resolution}
 The signature key can be obtained through "jwks_uri", see {{discovery}}.
 
@@ -2132,6 +2129,9 @@ multiple Receivers would lead to unintended data disclosure.
 }
 ~~~
 {: title="Example: SET with array 'aud' claim" #figarrayaud}
+
+### The "txn" claim {#txn-claim}
+Transmitters SHOULD set the "txn" claim value in Security Event Tokens (SETs). If the value is present, it MUST be unique to the underlying event that caused the Trasmitter to generate the Security Event Token (SET). The Transmitter, however, may use the same value in "txn" claim across different Security Events Tokens(SETs), such as session revoked and credential change, to indicate that the SETs originated from the same underlying cause or reason. 
 
 ### The "events" claim {#events-claim}
 The "events" claim SHOULD contain only one event. Multiple event type URIs are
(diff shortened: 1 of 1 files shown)


openid-sharedsignals-framework-1_0.md
```diff
@@ -2013,6 +2017,8 @@ The CAEP use cases that set the requirements are described in CAEP Use Cases (TO
 This section provides SSF profiling specifications for the Security Event Token (SET)
 {{RFC8417}} spec.
 
+{{RFC8417}} includes the usage of the `txn` claim in the spec. It is being called out here, to be used in a SET, as both a feedback loop and/or reconciling events between Transmitters and Receivers.
```
tulshi, line 2020: we need to use more normative language here, I think. E.g., "Transmitters SHOULD set the `txn` value in SETs. If present, this value MUST be unique to the underlying event that has caused the Transmitter to generate the SSF event."

iamseanodentity: Updated verbiage and called out differences from Transmitter underlying txn id's and how those can traverse different SET's

openid-sharedsignals-framework-1_0.md
```diff
@@ -2013,6 +2017,9 @@ The CAEP use cases that set the requirements are described in CAEP Use Cases (TO
 This section provides SSF profiling specifications for the Security Event Token (SET)
 {{RFC8417}} spec.
 
+{{RFC8417}} includes the usage of the `txn` claim in the specification. Transmitters SHOULD set the `txn` value in the Security Event Tokens (SETs). If the value is present, it MUST be unique to the underlying event that caused the Trasmitter to generate the Security Event Token (SET). The Transmitter, however, may use the same `txn` value across different Security Events Tokens(SETs), such as session revoked and credential change, to indicate that they originated from the same underlying cause or reason.
```
FragLegs, line 2020: The text looks good, but I wonder if it should be moved down to near section 10.2.2 or so, with the specifications about the `iss` claim, the `exp` claim, the `aud` claim, and the `events` claim?

FragLegs (changes requested):

openid-sharedsignals-framework-1_0.md
iamseanodentity: Open to it. I placed it where I did because this was the closest to the SET Profile. Atul, did say he liked the placement of it, so I am not married to the placement.

iamseanodentity: Done :)

openid-sharedsignals-framework-1_0.md
```diff
@@ -2123,6 +2130,9 @@ multiple Receivers would lead to unintended data disclosure.
 ~~~
 {: title="Example: SET with array 'aud' claim" #figarrayaud}
 
+### The "txn" claim {#txn-claim}
+Transmitters SHOULD set the "txn" claim value in Security Event Tokens (SETs). If the value is present, it MUST be unique to the underlying event that caused the Trasmitter to generate the Security Event Token (SET). The Transmitter, however, may use the same value in "txn" claim across different Security Events Tokens(SETs), such as session revoked and credential change, to indicate that the SETs originated from the same underlying cause or reason.
```
FragLegs, line 2134:
```suggestion
Transmitters SHOULD set the "txn" claim value in Security Event Tokens (SETs). If the value is present, it MUST be unique to the underlying event that caused the Transmitter to generate the Security Event Token (SET). The Transmitter, however, may use the same value in the "txn" claim across different Security Events Tokens (SETs), such as session revoked and credential change, to indicate that the SETs originated from the same underlying cause or reason. 
```

FragLegs (approved):

tulshi (approved):

PR #183, opened 2024-06-06
New session presented event

tulshi:


openid-caep-specification-1_0.md
```diff
@@ -758,12 +758,67 @@ The following is a non-normative example of the `session-established` event type
...
+ip
+: The array of IP addresses of the user as observed by the Transmitter. The value MUST be in the format of an array of strings, each one of which represents the RFC 4001 {{RFC4001}} string represetation of an IP address. (**NOTE**, this can be different from the one observed by the Receiver for the same user because of network translation)
+
+fp_ua
+: Fingerprint of the user agent computed by the Transmitter. (**NOTE**, this is not to identify the session, but to present some qualities of the session)
+
+ext_id
+: The external session identifier, which may be used to correlate this session with a broader session (e.g., a federated session established using SAML)
+
+risk_score
```
FragLegs, line 790: This doesn't belong here. If you want to convey a risk score then it should be sent in its own event type. Maybe a RiskScoreChanged event?

FragLegs (changes requested):

openid-caep-specification-1_0.md
timcappalli: I agree

openid-caep-specification-1_0.md
```diff
@@ -758,12 +758,67 @@ The following is a non-normative example of the `session-established` event type
...
+    "jti": "24c63fb56e5a2d77a6b512616ca9fa24",
+    "iat": 1615305159,
+    "aud": "https://sp.example.com/caep",
+    "sub_id": {
+      "format": "email",
+      "email": "someuser@somedomain.com"
+    },
+    "events": {
+        "https://schemas.openid.net/secevent/caep/event-type/session-presented": {
+          "ip": "192.168.1.12",
```
appsdesh, line 813: Array of IP addresses

iamseanodentity: Per the definition, yes: "The array of IP addresses of the user as observed by the Transmitter."

openid-caep-specification-1_0.md
iamseanodentity: I think placing a risk_score here makes sense as an optional claim being that the Transmitter can derive relative risk based on the identified characteristics.

openid-caep-specification-1_0.md
```diff
@@ -758,12 +758,67 @@ The following is a non-normative example of the `session-established` event type
...
+
+### Examples {#session-presented-examples}
+The following is a non-normative example of a Session Presented event:
+
+~~~json
+{
+    "iss": "https://idp.example.com/123456789/",
+    "jti": "24c63fb56e5a2d77a6b512616ca9fa24",
+    "iat": 1615305159,
+    "aud": "https://sp.example.com/caep",
```
iamseanodentity, line 806: add txn claim to the non-normative example

FragLegs:
When talking to @iamseanodentity I realized that I think the whole notion of "session presented" is problematic. All other SSF events tell us when something changes. But with session presented, the Receiver would need to compare each new event to stored previous events to determine if things like `risk_score`, `ip`, or `fp_ua` have changed.  It would be much more valuable to have RiskScoreChanged, IPChanged, and FingerprintChanged events.

tulshi:
> When talking to @iamseanodentity I realized that I think the whole notion of "session presented" is problematic. All other SSF events tell us when something changes. But with session presented, the Receiver would need to compare each new event to stored previous events to determine if things like `risk_score`, `ip`, or `fp_ua` have changed. It would be much more valuable to have RiskScoreChanged, IPChanged, and FingerprintChanged events.

A few things to consider:
1.  The event is important to build "Identity Threat Detection" capabilities. Such a capability can help receivers understand threats that span various transmitters. Each point-observation may not represent a threat, but together it might constitute one.
2. The "*changed" events you mention will not convey all possible threats because if an attacker uses the different IP addresses at different applications, but retains the same IP address for each application, then nothing will have changed at an application, and they would not send such "*changed" events.
3. For receivers who are not interested in such presence events (including those who find them difficult to process, like you describe), they can simply not subscribe to them, so unless you feel this event is not going to help further the CAEP charter, we should keep it.
4. Finally, the CAEP charter is to improve session security (paraphrasing), and since this event contributes to that charter, I believe it should be added.

openid-caep-specification-1_0.md
tulshi: updated this and the txn claim. Please review

openid-caep-specification-1_0.md
```diff
@@ -758,12 +758,60 @@ The following is a non-normative example of the `session-established` event type
...
+    "sub_id": {
+      "format": "email",
+      "email": "someuser@somedomain.com"
+    },
+    "events": {
+        "https://schemas.openid.net/secevent/caep/event-type/session-presented": {
+          "ip": ["192.168.1.12","10.1.1.1"]
+          "fp_ua": "abb0b6e7da81a42233f8f2b1a8ddb1b9a4c81611",
+          "ext_id": "12345",
+          "risk_score": 2,
```
appsdesh, line 809: This needs to be updated

openid-caep-specification-1_0.md
```diff
@@ -758,12 +758,60 @@ The following is a non-normative example of the `session-established` event type
...
+    "iat": 1615305159,
+    "aud": "https://sp.example.com/caep",
+    "txn": 8675309,
+    "sub_id": {
+      "format": "email",
+      "email": "someuser@somedomain.com"
+    },
+    "events": {
+        "https://schemas.openid.net/secevent/caep/event-type/session-presented": {
+          "ip": ["192.168.1.12","10.1.1.1"]
```
appsdesh, line 806: should we call this as `ips` (plural) to make it clear?

appsdesh (changes requested):

FragLegs (approved):
Looks good to me!

appsdesh (approved):
nit: I feel, AMR could be a list as well

timcappalli (approved):
I don't fully understand the use case / need for this event, but it is well formed and I don't think is harmful to include.

PR #190, opened 2024-06-26
fixed copyright in all files. Added security considerations to caep a…

tulshi:
…nd interop specs. Added notational conventions to interop spec


appsdesh (approved):

iamseanodentity (approved):
I did not think to look at the copyright date 🤦. It is also interesting, and I see why, how certain things need to be called out explicitly as such (i.e. SHOULD, SHALL, MAY)

timcappalli (commented):
LGTM

Issue #194, opened 2024-07-16
Fix markdown in "default_subjects" section


FragLegs:
The info is not rendering as a bulleted list because of missing newlines

FragLegs:
Need to split normative and non-normative change

Issue #186, opened 2024-06-17
Add "description" to list of fields that MAY be included in stream creation


FragLegs:
In section 7.1.1.1, we say that the stream creation request MAY include any Receiver-supplied fields, and then we list `events_requested` and `delivery`. We should either include `description` as well, or remove the list and let implementers discover which fields are Receiver-supplied above in section 7.1.1 (the latter option being more future-proof if we add any other Receiver-supplied fields in later versions).

tulshi:
Seems reasonable. You haven't marked this for v3, so would you like this to be addressed right now, or after v3?

FragLegs:
It's not essential. We can wait until after v3, since it is just a clarity edit.

FragLegs:
Pull this out of #195 into its own PR since it is a normative change

PR #195, opened 2024-08-12
Minor clarity edits

FragLegs:
Fixes:
- #194 
- #186

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -592,10 +592,12 @@ default_subjects
 > OPTIONAL. A string indicating the default behavior of newly created streams. If present,
   the value MUST be either "ALL" or "NONE". If not provided, the Transmitter behavior in
   this regard is unspecified.
+
 >  - "ALL" indicates that any subjects that are appropriate for the stream are added to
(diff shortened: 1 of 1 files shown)


openid-sharedsignals-framework-1_0.md
```diff
@@ -951,6 +953,7 @@ Configuration ({{stream-config}}) object:
 * `events_requested`
 * `delivery` : Note that in the case of the poll method, the `endpoint_url` value is
   supplied by the Transmitter.
+* `description`
```
tulshi, line 956: This line looks strange. Should there be an explanation after this?

FragLegs: It should not have an explanation, but I can move the note that is attached to `delivery` so that it is below the list.

timcappalli (approved):

openid-sharedsignals-framework-1_0.md
tulshi: I believe this is a normative change. Any reason to think of this as a non-normative change?

timcappalli:
This PR should split out normative and non-normative changes.

tulshi (approved):

Issue #201, opened 2024-09-10
A list of IPs as a new subject member in SSF


appsdesh:
As we are seeing the need for communicating IPs in more events, this issue proposes a list of IPs as a new [Complex Subject Member](https://openid.github.io/sharedsignals/openid-sharedsignals-framework-1_0.html#name-complex-subject-members) member.

The following events have IPs as members
- [Session Established (CAEP)](https://openid.github.io/sharedsignals/openid-caep-1_0.html#name-session-established)
- [Session Presented](https://openid.github.io/sharedsignals/openid-caep-1_0.html#name-session-presented)
- Newly proposed [Risk Level Change](https://github.com/openid/sharedsignals/issues/200)

Please refer [call notes from 9/10 for more context](https://hackmd.io/K9HVPrb0QvSFtp5sg7g0DA). cc: @FragLegs @iamseanodentity

FragLegs:
The members of a Complex Subject are all Simple Subjects. I think what we need is a new Simple Subject instead of a new Complex Subject field.

So a new IPs Simple Subject like
```
{
  "format": "ips",
  "ips": ["1.2.3.4", "2001:0000:130F:0000:0000:09C0:876A:130B"]
}
```

And then we could have a Complex Subject whose user field is that value, like
```
{
  "format": "complex",
  "user": {
    "format": "ips",
    "ips": ["1.2.3.4", "2001:0000:130F:0000:0000:09C0:876A:130B"]
  }
}
```

appsdesh:
@FragLegs - I am thinking more of declaring IPs as a first-class complex subject members along with existing `user`, `device`, `session`, `application`, `tenant`, `org_unit`, `group`.

With that, the simple subject would be -

```
{
  "format": "opaque",
  "ips": ["1.2.3.4", "2001:0000:130F:0000:0000:09C0:876A:130B"]
}
```

Complex would be -

```
{
  "format": "complex",
 "ips": {
    "format": "opaque",
    "ips": ["1.2.3.4", "2001:0000:130F:0000:0000:09C0:876A:130B"]
  }
}
```

If you are thinking of declaring `ips` as a `format` then we would need to go via adding it in [Security Events Identifiers Formats](https://www.iana.org/assignments/secevent/secevent.xhtml#:~:text=IETF-,%5BRFC8935%2C%20Section%202.4%5D,did)

I believe, that would need to be done outside SSWG

FragLegs:
[...]

PR #205, opened 2024-09-20
New CAEP event - Risk level change event

### LINKED ISSUE #200: New event to represent risk level changes
### Context 
A software vendor may deploy mechanisms to gather and analyze various signals associated with subjects such as users, devices, etc. These signals, which can originate from diverse channels and methods beyond the scope of this event description, are processed to derive an abstracted risk level representing the subject's current threat status.

The `risk-level-change` event is employed to communicate any modifications in a subject's assessed risk level between trusted entities. This event indicates whether the risk level has increased, decreased, or remained unchanged since the last assessment.

### Example Scenarios:

- Device Risk Assessment - A transmitter initially determines a device's risk-free status, assigning a LOW risk level. During a routine health scan, new software is detected. If the vendor assesses this software as legitimate but unauthorized, the risk level may be adjusted from LOW to MEDIUM. Conversely, if the software is identified as malicious and harmful, the risk level may be elevated from LOW to HIGH.
This event could be used in addition to the device-compliance-change when risk determination may/may not affect compliance policies. 
- User Risk Assessment - If multiple failed login attempts are detected, the user's risk level may increase to MEDIUM. Should it be discovered that the user’s password has been compromised, the risk may escalate to HIGH. Conversely, if the user subsequently resets their password and adopts high-assurance authentication methods, the risk level might be reduced.
This event could complement events like session-revoked. It could also help drive softer remediations by prompting for MFA, quarantining users, etc. 
- Tenant Risk Assessment - A tenant is under DDoS or has experienced a similar attack. The vendor wants to share this information with other entities.

Vendors may decide to ingest risks from various sources/providers, add weights to the risks calculated, and make their own policy determinations. The nonprescriptive nature of this event helps communicate the change in posture for the subject and leaves out the actions.

### Risk Level Classification:
The transmitter may utilize numeric ranges to categorize risk levels into predefined buckets such as LOW, MEDIUM, or HIGH. This classification helps in standardizing the risk communication and facilitating consistent risk management across different systems.

### [Event-Specific Claims](https://openid.github.io/sharedsignals/openid-caep-specification-1_0.html#name-event-specific-claims-4)

1. `current_level`  - REQUIRED, JSON string indicates the current level of the risk for the subject. Value MUST be one of LOW, MEDIUM, HIGH
2. `previous_level` - OPTIONAL, JSON string indicates the previously known level of the risk for the subject. Value MUST be one of LOW, MEDIUM, HIGH. If the Transmitter omits this value, the Receiver MUST assume that the previous assurance level is unknown to the Transmitter.
3. `ips` - OPTIONAL, The array of IP addresses of the subject as observed by the Transmitter. The value MUST be in the format of an array of strings, each one of which represents the RFC 4001 [RFC4001] string representation of an IP address. (NOTE, this can be different from the one observed by the Receiver for the same user because of network translation)

```
{
   "iss":"https://idp.example.com/3456789/",
   "jti":"07efd930f0977e4fcc1149a733ce7f78",
   "iat":1615305159,
   "aud":"https://sp.example2.net/caep",
   "events":{
      "https://schemas.openid.net/secevent/caep/event-type/risk-level-change":{
         "subject":{
            "user":{
               "format":"iss_sub",
               "iss":"https://idp.example.com/3456789/",
               "sub":"jane.smith@example.com"
            }
         },
         "current_level":"LOW",
         "previous_level":"HIGH",
         "ips":[
            "98.87.234.76"
         ],
         "event_timestamp":1615304991643,
         "reason_admin":{
            "en":"User's password detected in the pwned password dump"
         }
      }
   }
}
```

### Why CAEP event?
- Introducing a new profile will be a slow process
- CAEP has accommodated events that are not directly related to the session eg. device-compliance-change
- https://github.com/openid/sharedsignals/issues/158 proposes all the events to be in the JSON dictionary format, rather than in the spec. In the future, all the CAEP, RISC events could be converted to the JSON dictionaries. A use case could be conceptualized by creating profiles of the events.
- RISC profile is user account centric. This event is generic and could be applied to any subject

appsdesh:
New CAEP event - Risk level change event

Closes #200 

Removed `ips` claim from the event and created #201 to handle it separately for all CAEP events

### DIFF
--- a/openid-caep-1_0.md
+++ b/openid-caep-1_0.md
@@ -822,6 +822,62 @@ The following is a non-normative example of a Session Presented event:
     }}
 ~~~
 
+## Risk Level Change {#risk-level-change}
+Event Type URI:
+
+`https://schemas.openid.net/secevent/caep/event-type/risk-level-change`
+
+A vendor may deploy mechanisms to gather and analyze various signals associated with subjects such as users, devices, etc. These signals, which can originate from diverse channels and methods beyond the scope of this event description, are processed to derive an abstracted risk level representing the subject's current threat status.
+
+The Risk Level Change event is employed by the Transmitter to communicate any modifications in a subject's assessed risk level at the time indicated by the `event_timestamp` field in the Risk Level Change event. The Transmitter may generate this event to indicate:
+
+* User's risk has changed due to potential suspecious access from unknown destination, password compromise, addition of strong authenticator or other reasons.
+* Device's risk has changed due to installation of unapproved software, connection to insecure pheripheral device, encryption of data or other reasons.
+* Any other subject's risk changes due to variety of reasons.
+
+
+### Event Specific Claims {#risk-level-change-event-specific-claims}
+
+risk_reason
+: RECOMMENDED, JSON string: indicates the reason that contributed to the risk level changes by the Transmitter.
+
+principal
+: REQUIRED, JSON string: representing the principal entity involved in the observed risk event, as identified by the transmitter. The subject principal can be one of the following entities USER, DEVICE, SESSION, TENANT, ORG_UNIT, GROUP, or any other entity as defined in {{Section 2 of SSF}}. This claim identifies the primary subject associated with the event, and helps to contextualize the risk relative to the entity involved.
+
+current_level 
+: REQUIRED, JSON string: indicates the current level of the risk for the subject. Value MUST be one of LOW, MEDIUM, HIGH
+
+previous_level 
+: OPTIONAL, JSON string: indicates the previously known level of the risk for the subject. Value MUST be one of LOW, MEDIUM, HIGH. If the Transmitter omits this value, the Receiver MUST assume that the previous risk level is unknown to the Transmitter.
+
+
+### Examples {#risk-level-change-examples}
+The following is a non-normative example of a Risk Level Change event:
+
+~~~json
+{
+    "iss": "https://idp.example.com/123456789/",
+    "jti": "24c63fb56e5a2d77a6b512616ca9fa24",
+    "iat": 1615305159,
+    "aud": "https://sp.example.com/caep",
+    "txn": 8675309,
+    "sub_id": {
+      "format": "iss_sub",
+      "iss": "https://idp.example.com/3456789/",
+      "sub": "jane.doe@example.com"
+    },
+    "events":{
+      "https://schemas.openid.net/secevent/caep/event-type/risk-level-change":{
+         "current_level": "LOW",
+         "previous_level": "HIGH",
+         "event_timestamp": 1615304991643,
+         "principal": "USER",
+         "risk_reason": "PASSWORD_FOUND_IN_DATA_BREACH"
+      }
+   }
+}
+~~~
+
 # Security Considerations
 Any implementations of events described in this document SHOULD comply with the Shared Signals Framework {{SSF}}. Exchanging events described herein without complying with the Shared Signals Framework {{SSF}} may result in security issues.
 
--- a/openid-caep-1_0.md
+++ b/openid-caep-1_0.md
@@ -735,7 +735,7 @@ The `event_timestamp` in this event type specifies the time at which the session
 The following optional claims MAY be included in the Session Established event:
 
 ips
-: The array of IP addresses of the user as observed by the Transmitter. The value MUST be in the format of an array of strings, each one of which represents the RFC 4001 {{RFC4001}} string represetation of an IP address. (**NOTE**, this can be different from the one observed by the Receiver for the same user because of network translation)
+: The array of IP addresses of the user as observed by the Transmitter. The value MUST be in the format of an array of strings, each one of which represents the RFC 4001 {{RFC4001}} string representation of an IP address. (**NOTE**, this can be different from the one observed by the Receiver for the same user because of network translation)
 
 fp_ua
 : Fingerprint of the user agent computed by the Transmitter. (**NOTE**, this is not to identify the session, but to present some qualities of the session)
@@ -790,7 +790,7 @@ The Session Presented event signifies that the Transmitter has observed the sess
 The following optional claims MAY be present in a Session Presented event:
 
 ips
-: The array of IP addresses of the user as observed by the Transmitter. The value MUST be in the format of an array of strings, each one of which represents the RFC 4001 {{RFC4001}} string represetation of an IP address. (**NOTE**, this can be different from the one observed by the Receiver for the same user because of network translation)
+: The array of IP addresses of the user as observed by the Transmitter. The value MUST be in the format of an array of strings, each one of which represents the RFC 4001 {{RFC4001}} string representation of an IP address. (**NOTE**, this can be different from the one observed by the Receiver for the same user because of network translation)
 
 fp_ua
 : Fingerprint of the user agent computed by the Transmitter. (**NOTE**, this is not to identify the session, but to present some qualities of the session)
--- a/openid-caep-1_0.md
+++ b/openid-caep-1_0.md
@@ -822,6 +822,7 @@ The following is a non-normative example of a Session Presented event:
     }}
 ~~~
 
+
 ## Risk Level Change {#risk-level-change}
 Event Type URI:
 
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -436,23 +436,28 @@ The following are hypothetical examples of SETs that conform to the Shared Signa
   "txn": 8675309,
   "aud": "636C69656E745F6964",
   "sub_id": {
-      "format": "complex",
-      "user": {
-          "format": "iss_sub",
-          "iss": "https://idp.example.com/3957ea72-1b66-44d6-a044-d805712b9288/",
-          "sub": "jane.smith@example.com"
-      },
-      "device": {
-          "format": "iss_sub",
-          "iss": "https://idp.example.com/3957ea72-1b66-44d6-a044-d805712b9288/",
-          "sub": "e9297990-14d2-42ec-a4a9-4036db86509a"
-      }
+    "format": "complex",
+    "user": {
+      "format": "iss_sub",
+      "iss": "https://idp.example.com/3957ea72-1b66-44d6-a044-d805712b9288/",
+      "sub": "jane.smith@example.com"
+    },
+    "device": {
+      "format": "iss_sub",
+      "iss": "https://idp.example.com/3957ea72-1b66-44d6-a044-d805712b9288/",
+      "sub": "e9297990-14d2-42ec-a4a9-4036db86509a"
+    }
   },
   "events": {
     "https://schemas.openid.net/secevent/caep/event-type/session-revoked": {
       "initiating_entity": "policy",
-      "reason_admin": "Policy Violation: C076E82F",
-      "reason_user": "Landspeed violation.",
+      "reason_admin": {
+        "en": "Policy Violation: C076E82F"
+      },
+      "reason_user": {
+        "en": "Land speed violation.",
+        "es": "Violación de velocidad en tierra."
+      },
       "event_timestamp": 1600975810
     }
   }


openid-caep-1_0.md
```diff
@@ -822,6 +822,61 @@ The following is a non-normative example of a Session Presented event:
...
 ~~~
 
+## Risk Level Change {#risk-level-change}
+Event Type URI:
+
+`https://schemas.openid.net/secevent/caep/event-type/risk-level-change`
+
+A vendor may deploy mechanisms to gather and analyze various signals associated with subjects such as users, devices, etc. These signals, which can originate from diverse channels and methods beyond the scope of this event description, are processed to derive an abstracted risk level representing the subject's current threat status.
+
+The Risk Level Change event is employed  by the Trasnmitter to communicate any modifications in a subject's assessed risk level at the time indicated by the `event_timestamp` field in the Risk Level Change event. The Transmitter may generate this event to indicate:
```
FragLegs, line 832:
```suggestion
The Risk Level Change event is employed by the Transmitter to communicate any modifications in a subject's assessed risk level at the time indicated by the `event_timestamp` field in the Risk Level Change event. The Transmitter may generate this event to indicate:
```

openid-caep-1_0.md
```diff
@@ -822,6 +822,61 @@ The following is a non-normative example of a Session Presented event:
...
+
+* User's risk has changed due to potential suspecious access from unknown destination, password compromise, addition of strong authenticator or other reasons.
+* Device's risk has changed due to installation of unapproved software, connection to insecure pheripheral device, encryption of data or other reasons.
+* Any other subject's risk changes due to variety of reasons.
+
+
+### Event Specific Claims {#risk-level-change-event-specific-claims}
+The following optional claims MAY be present in a Session Presented event:
+
+reason_admin
```
FragLegs, line 842: I think it is a mistake to make this optional field required and to use it for passing important info about the event.

openid-caep-1_0.md
```diff
@@ -822,6 +822,61 @@ The following is a non-normative example of a Session Presented event:
...
+The Risk Level Change event is employed  by the Trasnmitter to communicate any modifications in a subject's assessed risk level at the time indicated by the `event_timestamp` field in the Risk Level Change event. The Transmitter may generate this event to indicate:
+
+* User's risk has changed due to potential suspecious access from unknown destination, password compromise, addition of strong authenticator or other reasons.
+* Device's risk has changed due to installation of unapproved software, connection to insecure pheripheral device, encryption of data or other reasons.
+* Any other subject's risk changes due to variety of reasons.
+
+
+### Event Specific Claims {#risk-level-change-event-specific-claims}
+The following optional claims MAY be present in a Session Presented event:
+
```
FragLegs, line 841: Could we add a field that describes the type of risk? You mention in your example user risk and device risk. What if we had a field called `risk_type` whose values were an enum:
- DEVICE
- USER
- ACCOUNT
- OTHER

There may be other values that we should include in this enum as well.

FragLegs (changes requested):

openid-caep-1_0.md
appsdesh: @FragLegs @timcappalli we could define a new attribute that describes this, please let me know if you have any other suggestions

openid-caep-1_0.md
appsdesh: @FragLegs I thought we could drive the type using the subject claim. With a single enum it would be hard to flag a single risk event if it applies to more than 1 entity like USER + DEVICE etc.

openid-caep-1_0.md
```diff
@@ -822,6 +822,61 @@ The following is a non-normative example of a Session Presented event:
...
+The Risk Level Change event is employed by the Transmitter to communicate any modifications in a subject's assessed risk level at the time indicated by the `event_timestamp` field in the Risk Level Change event. The Transmitter may generate this event to indicate:
+
+* User's risk has changed due to potential suspecious access from unknown destination, password compromise, addition of strong authenticator or other reasons.
+* Device's risk has changed due to installation of unapproved software, connection to insecure pheripheral device, encryption of data or other reasons.
+* Any other subject's risk changes due to variety of reasons.
+
+
+### Event Specific Claims {#risk-level-change-event-specific-claims}
+The following optional claims MAY be present in a Session Presented event:
```
ysarig75, line 840: Typo: "in a Session Presented event" (it should be "in a Risk Level Change event")

openid-caep-1_0.md
```diff
@@ -822,6 +822,61 @@ The following is a non-normative example of a Session Presented event:
...
+
+reason_admin
+: REQUIRED, JSON string indicates the reason that contributed to the risk level changes by the Transmitter. Overall this is an OPTIONAL claim, but marked as REQUIRED for Risk Level Change to indicate reasons behind the risk level change.
+
+current_level 
+: REQUIRED, JSON string indicates the current level of the risk for the subject. Value MUST be one of LOW, MEDIUM, HIGH
```
ysarig75, line 846: Mandating the use of a LOW, MEDIUM, HIGH for risk levels may be too restricted for some usage. Some Companies may want to use a numeric range (e.g., 1-5 or 1-10) instead without needing to map it into LOW, MEDIUM, HIGH range.

appbugs:
The Risk level change event can be used by implementers as a "catch all" event. For example, an receiver could decide to use Risk level change instead of implementing the CAEP events. If we have a Risk level change event, which is prescriptive, we'd allow implementers to go the easy route and implement ONLY Risk level change event. Is that a behavior we want to encourage?

As we discussed today CAEP events are descriptive, while Risk event is prescriptive. I don't see prescriptive and descriptive events coexist in the same spec. My suggestion is to wait with adding this event to the final spec.

appsdesh:
> The Risk level change event can be used by implementers as a "catch all" event. For example, an receiver could decide to use Risk level change instead of implementing the CAEP events. If we have a Risk level change event, which is prescriptive, we'd allow implementers to go the easy route and implement ONLY Risk level change event. Is that a behavior we want to encourage?
> 
> As we discussed today CAEP events are descriptive, while Risk event is prescriptive. I don't see prescriptive and descriptive events coexist in the same spec. My suggestion is to wait with adding this event to the final spec.

This event is the same as any other event in the CAEP spec, and it is NO WAY prescriptive. The Reciver may decide to do NOTHING on the risk being HIGH, same way as they chose to ingnore existing Session Revoked or Credential Change events. 

Whether and how organizations respond to any of the CAEP events is completely dependent on their risk tolerance levels.

appbugs:
> #200



> > The Risk level change event can be used by implementers as a "catch all" event. For example, an receiver could decide to use Risk level change instead of implementing the CAEP events. If we have a Risk level change event, which is prescriptive, we'd allow implementers to go the easy route and implement ONLY Risk level change event. Is that a behavior we want to encourage?
> > As we discussed today CAEP events are descriptive, while Risk event is prescriptive. I don't see prescriptive and descriptive events coexist in the same spec. My suggestion is to wait with adding this event to the final spec.
> 
> This event is the same as any other event in the CAEP spec, and it is NO WAY prescriptive. The Reciver may decide to do NOTHING on the risk being HIGH, same way as they chose to ingnore existing Session Revoked or Credential Change events.
> 
> Whether and how organizations respond to any of the CAEP events is completely dependent on their risk tolerance levels. The event can describe what led to Tx raising or reducing the risk and description around that, this is very flexible mechanism to share the risks with vendors.

The Risk level change event is not prescribe what to do, it just augments different security events into the Risk level change event. The question is do we need to have an event that factors other CAEP events to come up with Low/Medium/High or a Tx can just use the existing events. If there are other events that Risk level change is factoring, can we just add those events as stand-alone to the CAEP spec?

appsdesh:
> > #200
> 
> > > The Risk level change event can be used by implementers as a "catch all" event. For example, an receiver could decide to use Risk level change instead of implementing the CAEP events. If we have a Risk level change event, which is prescriptive, we'd allow implementers to go the easy route and implement ONLY Risk level change event. Is that a behavior we want to encourage?
> > > As we discussed today CAEP events are descriptive, while Risk event is prescriptive. I don't see prescriptive and descriptive events coexist in the same spec. My suggestion is to wait with adding this event to the final spec.
> > 
> > 
> > This event is the same as any other event in the CAEP spec, and it is NO WAY prescriptive. The Reciver may decide to do NOTHING on the risk being HIGH, same way as they chose to ingnore existing Session Revoked or Credential Change events.
> > Whether and how organizations respond to any of the CAEP events is completely dependent on their risk tolerance levels. The event can describe what led to Tx raising or reducing the risk and description around that, this is very flexible mechanism to share the risks with vendors.
> 
> The Risk level change event does not prescribe what to do, it just augments different security events into the Risk level change event. The question is do we need to have an event that factors other CAEP events to come up with Low/Medium/High or a Tx can just use the existing CAEP events. If there are other events that Risk level change is factoring, can we just add those events as stand-alone events to the CAEP spec?

As discussed in the call, risk indicators are a function of the Tx maturity. Risk calculations are outside the scope of this specification. Risk could encompass things described outside CAEP and RISK specs. You are free to propose any events that describe your usecases. 

This event is standalone and represents risk as a first-class object to be shared with the co-operating parties.

openid-caep-1_0.md
appsdesh: Added `risk_reason` as a new field to describe the nature of the risk

openid-caep-1_0.md
appsdesh: [Assurance level change](https://openid.net/specs/openid-caep-1_0-ID2.html#section-3.4.1-1.2.1) uses `namespace`, but in that example, there are standardized ways of representing levels.

In this case, I feel even the numeric system will be subjective, it will not help bring standardization. 

Another option is to plainly define a `score` as an optional attribute that could satisfy your use case, but I believe that could be solved by adopting https://github.com/openid/sharedsignals/issues/221

tulshi: The [assurance level change](https://openid.net/specs/openid-caep-1_0-ID2.html#name-assurance-level-change) event has a "Any other value..." escape, that makes it possible to define your own namespace.  I feel having a standard set of values - whether Low, Medium, High or 1-10, will improve interoperability, so I'm in favor of keeping it simple. Like Apoorva says, we can include additional information in the "event data" field. I feel the numeric range is slightly more expressive, and you could define LOW, MEDIUM and HIGH as three points in that range, so it subsumes that.

openid-caep-1_0.md
appsdesh: @tulshi I described that the existing subject of any CAEP event describes who this event is for, and it is clarified in [this](https://openid.github.io/sharedsignals/openid-sharedsignals-framework-1_0.html#name-subject-members) section. In other words, if the risk level change event has 2 subjects, a user and a device, then it should be perceived as risk being associated with both entities.

tulshi: I guess my question is let's say there is a user X, who has a phone number Y. The transmitter wants to send out a risk level change about the user (and not their phone), but wants to describe the user using their phone number (i.e. the subject format is the ["phone_number"](https://www.rfc-editor.org/rfc/rfc9493.html#name-phone-number-identifier-for)), then how can the Transmitter convey this information?

appsdesh: Thanks, @tulshi, for your comment.

I think this could also be a similar situation for session-related events.

e.g. There are separate sessions created with the user as a principal and the device as a principal.

> The transmitter wants to send out a session revoked/presented/established about the user (and not their device), but wants to describe the user using their phone number (i.e. the subject format is the ["phone_number"](https://www.rfc-editor.org/rfc/rfc9493.html#name-phone-number-identifier-for))

openid-caep-1_0.md
```diff
@@ -822,6 +822,60 @@ The following is a non-normative example of a Session Presented event:
...
+
+* User's risk has changed due to potential suspecious access from unknown destination, password compromise, addition of strong authenticator or other reasons.
+* Device's risk has changed due to installation of unapproved software, connection to insecure pheripheral device, encryption of data or other reasons.
+* Any other subject's risk changes due to variety of reasons.
+
+
+### Event Specific Claims {#risk-level-change-event-specific-claims}
+
+risk_reason
+: REQUIRED, JSON string indicates the reason that contributed to the risk level changes by the Transmitter.
```
FragLegs, line 842:
```suggestion
: REQUIRED, JSON object: a localizable message indicating the reason that contributed to the risk level changes by the Transmitter. The object MUST contain one or more key/value pairs, with a BCP47 {{RFC5646}} language tag as the key and the locale-specific administrative message as the value.
```

FragLegs: Why is this REQUIRED? It seems like it ought to be an optional field

openid-caep-1_0.md
```diff
@@ -822,6 +822,60 @@ The following is a non-normative example of a Session Presented event:
...
+
+current_level 
+: REQUIRED, JSON string indicates the current level of the risk for the subject. Value MUST be one of LOW, MEDIUM, HIGH
```
FragLegs, line 845:
```suggestion
: REQUIRED, string: indicates the current level of the risk for the subject. Value MUST be one of LOW, MEDIUM, HIGH
```

openid-caep-1_0.md
```diff
@@ -822,6 +822,60 @@ The following is a non-normative example of a Session Presented event:
...
+
+previous_level 
+: OPTIONAL, JSON string indicates the previously known level of the risk for the subject. Value MUST be one of LOW, MEDIUM, HIGH. If the Transmitter omits this value, the Receiver MUST assume that the previous assurance level is unknown to the Transmitter.
```
FragLegs, line 848:
```suggestion
: OPTIONAL, string: indicates the previously known level of the risk for the subject. Value MUST be one of LOW, MEDIUM, HIGH. If the Transmitter omits this value, the Receiver MUST assume that the previous assurance level is unknown to the Transmitter.
```

FragLegs: I think the second statement ought to be a SHOULD:

> If the Transmitter omits this value, the Receiver SHOULD assume that the previous assurance level is unknown to the Transmitter.

openid-caep-1_0.md
FragLegs: A `principal` claim works for me!

openid-caep-1_0.md
appsdesh: Following the same convention of the other CAEP events

openid-caep-1_0.md
appsdesh: Following the same convention of the other CAEP events

openid-caep-1_0.md
appsdesh: `risk_reason` is marked required since, without this, there is no benefit for the receiver to process such an event. Transmitter updating the risk will be a black box process and hinder the visibility.

Also, it may not make sense to localize the message as this is intended for the receiver to make sense why Tx decided to update the risk. This could become ENUM between Tx-Rx. (Happy to update the description if that clarifies)

User-visible messages would be part of `reason_admin` and `reason_user` which could be localized

openid-caep-1_0.md
```diff
@@ -822,6 +822,63 @@ The following is a non-normative example of a Session Presented event:
...
+      "format": "iss_sub",
+      "iss": "https://idp.example.com/3456789/",
+      "sub": "jane.doe@example.com"
+    },
+    "events":{
+      "https://schemas.openid.net/secevent/caep/event-type/risk-level-change":{
+         "current_level": "LOW",
+         "previous_level": "HIGH",
+         "event_timestamp": 1615304991643,
+         "risk_reason":{
```
FragLegs, line 874: This should be a string, not be a dict of language to string

openid-caep-1_0.md
```diff
@@ -822,6 +822,63 @@ The following is a non-normative example of a Session Presented event:
...
+    "aud": "https://sp.example.com/caep",
+    "txn": 8675309,
+    "sub_id": {
+      "format": "iss_sub",
+      "iss": "https://idp.example.com/3456789/",
+      "sub": "jane.doe@example.com"
+    },
+    "events":{
+      "https://schemas.openid.net/secevent/caep/event-type/risk-level-change":{
+         "current_level": "LOW",
```
FragLegs, line 871: Can we add "principal" to the example?

FragLegs (approved):

openid-caep-1_0.md
```diff
@@ -822,6 +822,62 @@ The following is a non-normative example of a Session Presented event:
...
+
+### Event Specific Claims {#risk-level-change-event-specific-claims}
+
+risk_reason
+: RECOMMENDED, JSON string: indicates the reason that contributed to the risk level changes by the Transmitter.
+
+principal
+: REQUIRED, JSON string: representing the principal entity involved in the observed risk event, as identified by the transmitter. The subject principal can be one of the following entities USER, DEVICE, SESSION, TENANT, ORG_UNIT, GROUP, or any other entity as defined in {{Section 2 of SSF}}. This claim identifies the primary subject associated with the event, and helps to contextualize the risk relative to the entity involved.
+
+current_level
```
tulshi, line 847: @appsdesh any thoughts on changing this to a numeric value on a fixed scale (e.g. 10)? We discussed this during the [SSWG meeting on 11/5](https://hackmd.io/@oidf-wg-sse/wg-meeting-20241105), but I don't remember if we agreed to switch to a numeric value. three of those values could map to "low", "medium" and "high" as you require, but it will give a little more expressivity.

appsdesh: @tulshi welcome back! we already aligned on it in one of the comments [here](https://github.com/openid/sharedsignals/pull/205#discussion_r1843240302).

tulshi (approved):

appsdesh:
Continuning onto https://github.com/openid/sharedsignals/pull/231
(The branch was messed up due to merge issues and limited github privileges)

Issue #158, opened 2024-04-25
Explore machine reachable approach to event definitions


timcappalli:
We've long discussed the challenges with the event specs, which are effectively just dictionaries, being formal specifications that go through full OIDF process. This can make it challenging to add new events quickly which can make the effort feel less accessible. 

This issue is to track and explore a more machine readable approach to defining events, such as using a JSON schema file mainted on GitHub with version controls.

Some early thoughts:

- There would no longer be a single document 
- Events would be versioned instead of the document (likely using a path in the event URI)
- The event URI would be dereferencable to the JSON schema (or other file)

FragLegs:
I feel strongly that we should _not_ include the entire SET in the definition. Just the event itself. Since the `events` claim of the SET is a dictionary mapping these URNs to objects, it will be easier to model if we have schemas for each individual object. Also, if the schema included the full SET, then any changes to the SET structure (like the proposed inclusion of `txn`) would force us to update all of the event schemas.

tulshi:
A few comments / observations:

1. If it's not a single document, how will all such documents be organized?
2. The initial issue talks about event versioning, but the detailed description does not indicate how the versioning will be captured.
3. How do you describe the optionality of specific fields in the current proposal?
4. We need to be able to give non-normative examples of such events. We need to think of how to specify that in this format
5. We need semantics around normative and non-normative language in the description of the fields.
6. What is the OpenID Foundation process that we are proposing for using this format? How does someone propose it, how does the WG discuss / work on it, when is OIDF required to review it (like in the case of implementer's draft)?

timcappalli:
@timcappalli to add RISC example

timcappalli:
From 2024-05-28 in person discussion:

- Sentiment is generally positive for this
- @timcappalli to add a RISC example to this issue
- Need to define a new spec (ex: "Shared Signals Event Definition Specification") that explains the schema, process, etc (this spec would replace RISC and CAEP documents)
- Need to check OIDF policy document on whether the schema files are normative or non-normative in the context of IPR as those checks are done during spec process (ID, etc)

tulshi:
Notes from the call on 11/19:

- Jen reviewed the current proposal
- Mike Jones had some concerns about how this would work?
  - He thought a JSON schema won't be sufficient, and a spec would be required
  - (Jen) So does the WG think that just the schema is sufficient or we need a spec.
  - (Atul) What's an example of the insufficiency?
[...]

PR #209, opened 2024-09-24
First iteration of Event Def Spec

jischr:
This PR addresses https://github.com/openid/sharedsignals/issues/158

### DIFF
--- a/openid-sharedsignals-eventdef-specification-1_0.md
+++ b/openid-sharedsignals-eventdef-specification-1_0.md
@@ -2,7 +2,7 @@
 title: OpenID Shared Signals Event Definition Specification
 abbrev: SSFEvents-Spec
 docname: openid-sharedsignals-eventdef-specification-1_0
-date: 2024-06-06
+date: 2024-10-22
 
 ipr: none
 cat: std
@@ -110,9 +110,9 @@ The Shared Signals Framework {{SSF}} enables sharing of signals and events betwe
 This specification defines how to translate normative SSF event requirements, such as from CAEP and RISC events, into a JSON Schema. {{JSONSchema}} is a standardized way to describe the structure, constraints, and data types within a JSON document. JSON Schemas are often used with {{JSONSchemaValidation}} to automatically check if a JSON document adheres to the defined schema, thus ensuring data integrity.
 
 Using JSON Schema to describe SSF has the following benefits:
-- Machine readability for the auto generation of stubs
 - Faster process to create, update and get approval for new event types
 - Easy and appropriate format to describe event types, as opposed to traditional specification text
+- Machine readability for creating and validating SSF event paylods
 - Independent versioning to reduce the friction between the SSF core specification and event type publications
 - Allows new SSF events to be incorporated into the SSF ecosystem
 
@@ -130,7 +130,7 @@ The schema is made up of the following top-level JSON keys:
 : REQUIRED, JSON string: URI of the version of JSON Schema that this document adheres to. E.g., `https://json-schema.org/draft/2020-12/schema`. Normative Schema Keyword.
 
 \\$id
-: REQUIRED, JSON string: Event Identifier of the SET and URI of the schema. MUST be publicly accessible or available out-of-band and resolve to JSON Schema document. Normative Schema Keyword.
+: REQUIRED, JSON string: URI of the schema, also known as the Event Identifier of the SET appended in accordance with[Schema Ids](#schema-ids). MUST be publicly accessible or available out-of-band and resolve to JSON Schema document. Normative Schema Keyword.
 
 title
 : REQUIRED, JSON string: Title of the schema. Informational Schema Annotation.
@@ -149,7 +149,7 @@ The following is a non-normative example of the top level properties of a JSON S
 ~~~ json
 {
    "$schema":"https://json-schema.org/draft/2020-12/schema",
-   "$id":"https://schemas.openid.net/secevent/caep/event-type/session-revoked/v1.schema.json",
+   "$id":"https://schemas.openid.net/secevent/caep/event-type/session-revoked/1.0.0/schema.json",
    "title":"Session Revoked",
    "description":"Session Revoked signals that the session identified by the subject has been revoked. The explicit session identifier may be directly referenced in the subject or other properties of the session may be included to allow the receiver to identify applicable sessions.",
    "type":"object",
@@ -198,7 +198,7 @@ The following is a non-normative example of the top level properties of a JSON S
 ~~~ json
 {
    "$schema":"https://json-schema.org/draft/2020-12/schema",
-   "$id":"https://schemas.openid.net/secevent/caep/event-type/session-revoked/v1.schema.json",
+   "$id":"https://schemas.openid.net/secevent/caep/event-type/session-revoked/1.0.0/schema.json",
    "title":"Session Revoked",
    "description":"Session Revoked signals that the session identified by the subject has been revoked. The explicit session identifier may be directly referenced in the subject or other properties of the session may be included to allow the receiver to identify applicable sessions.",
    "type":"object",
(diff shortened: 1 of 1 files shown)


openid-sharedsignals-eventdef-specification-1_0.md
```diff
@@ -0,0 +1,282 @@
...
+
+SSF Implementers may find that existing registered SSF event types do not meet the needs of their applications. In that case, they may propose a new SSF event type and register its schema. To do so, an implementer MUST create a request in the form of a Pull Request to https://github.com/openid/sharedsignals and meet the following requirements. The pull request will be reviewed by the Shared Signals Working Group and accepted at their discression.
+
+1. Author(s) of the pull request MUST be at least a contributing member of the OpenID Foundation.
+1. The Pull Request MUST contain a human readable description of the new SSF event type.
+1. The $id of the schema MUST be publicly accesbile on the internet and resolve to the schema document.
+1. The "title" and "description" of the schema must be meaningful and indicative of its function.
+1. The naming of all schema properties MUST be indicative of its function.
+1. Schemas may not be removed, only deprecated. Any changes to schemas must follow semantic versioning.
+
```
appsdesh, line 266: Should we specify JSON schema for the SET and call out all the mandatory claims of the SET are required? This will help developers to generate entire container objects, rather than only the sub-object `events`

openid-sharedsignals-eventdef-specification-1_0.md
```diff
@@ -0,0 +1,282 @@
...
+    target: http://openid.net/specs/openid-connect-core-1_0.html#IDToken
+    title: OpenID Connect Core 1.0 - ID Token
+  RFC2119:
+  RFC8174:
+  RFC5280:
+  RFC5646:
+  RFC6711:
+  RFC8176:
+  SSF:
+    target: http://openid.net/specs/openid-sse-framework-1_0.html
```
FragLegs, line 57: This is referencing a very out of date draft. Please update to draft 3: https://openid.net/specs/openid-sharedsignals-framework-1_0-03.html

FragLegs (changes requested):

openid-sharedsignals-eventdef-specification-1_0.md
```diff
@@ -0,0 +1,282 @@
...
+Using JSON Schema to describe SSF has the following benefits:
+- Allows machine readability to auto generation of stubs.
+- It enables a faster process to create, update and get approval for new event types. 
+- JSON schema, rather than spec texts, is a more appropriate format to describe event types. 
+- It also allows event types to be versioned independently thus reducing the friction between the SSF core specification and event type publication.
+- Allows more events to be incorporated easily in the standards space beyond the usecases and charters of CAEP, RISC etc.
+
+# JSON Schema Defintion
+
+The following section describes how a map a SSF event to a JSON Schema Document.
```
timcappalli, line 126:
```suggestion
The following section describes how a map a SSF event to a JSON Schema document.
```

openid-sharedsignals-eventdef-specification-1_0.md
```diff
@@ -0,0 +1,282 @@
...
+
+As defined in {{Section 4.3 of JSON Schema}}, a JSON Schema document, also called a "schema", is a JSON document used to describe another JSON Document, known as an instance.
```
timcappalli, line 128: change for @FragLegs comment: 

```suggestion
As defined in Section 4.3 of {{JSON Schema}}, a JSON Schema document, also called a "schema", is a JSON document used to describe another JSON Document, known as an instance.
```

openid-sharedsignals-eventdef-specification-1_0.md
```diff
@@ -0,0 +1,282 @@
...
+    },
+    "required": [ "initiating_entity" ]
+   }
+}
+~~~
+# Registry
+
+This section serves as a registry for the schemas of all registered SSF Event Types.
+
+| Event Type | Schema URI | Description
```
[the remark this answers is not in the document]
timcappalli: My goal with this effort was to eliminate the need for this classification. An event should be descriptive enough to not need a bucket.

openid-sharedsignals-eventdef-specification-1_0.md
```diff
@@ -0,0 +1,282 @@
...
+
+properties
+: REQUIRED JSON object: Object that defines the keys in the JSON data that are being validated. Normative Validation Keyword.
+
+
+The following is a non-normative example of a JSON Schema document for a Session Revoked CAEP event.
+~~~ json
+{
+   "$schema":"https://json-schema.org/draft/2020-12/schema",
+   "$id":"https://schemas.openid.net/secevent/caep/event-type/session-revoked/v1.schema.json",
```
timcappalli, line 158: I think we might need to define a mapping of the event identifier in the SET to the JSON schema identifier.

E.g. the event identifier is `$id` with `.schema.json` stripped?

And then we can say the inverse: an event's schema can be found by appending `.schema.json` to the event identifier and then dereferencing it.

openid-sharedsignals-eventdef-specification-1_0.md
```diff
@@ -0,0 +1,282 @@
...
+}
+~~~
+# Registry
+
+This section serves as a registry for the schemas of all registered SSF Event Types.
+
+| Event Type | Schema URI | Description
+|------|-------------|
+| CAEP  | <schema uri here> | brief description |
+| RISC  |  <schema uri here>  | brief description |
```
timcappalli, line 252: hoping we can eliminate the need to classify things as "CAEP" or "RISC". If it needs to be done, it can be done via the event URI.

openid-sharedsignals-eventdef-specification-1_0.md
```diff
@@ -0,0 +1,282 @@
...
+
+{: title="Registered Event Types" #eventtypestable}
+
+## The Registration Process
+
+SSF Implementers may find that existing registered SSF event types do not meet the needs of their applications. In that case, they may propose a new SSF event type and register its schema. To do so, an implementer MUST create a request in the form of a Pull Request to https://github.com/openid/sharedsignals and meet the following requirements. The pull request will be reviewed by the Shared Signals Working Group and accepted at their discression.
+
+1. Author(s) of the pull request MUST be at least a contributing member of the OpenID Foundation.
+1. The Pull Request MUST contain a human readable description of the new SSF event type.
+1. The $id of the schema MUST be publicly accesbile on the internet and resolve to the schema document.
```
timcappalli, line 262: An alternative is that we use issues with issue templates, so we can require certain information be provided.

timcappalli (changes requested):
Initial pass before today's call.

openid-sharedsignals-eventdef-specification-1_0.md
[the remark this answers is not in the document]
jischr:
```suggestion
As defined in Section 4.3 of {{JSONSchema}}, a JSON Schema document, also called a "schema", is a JSON document used to describe another JSON Document, known as an instance.
```

jischr: Thanks! must have missed this one.

openid-sharedsignals-eventdef-specification-1_0.md
jischr: i think thats a good idea.

openid-sharedsignals-eventdef-specification-1_0.md
```diff
@@ -0,0 +1,282 @@
...
+1. The "title" and "description" of the schema must be meaningful and indicative of its function.
+1. The naming of all schema properties MUST be indicative of its function.
+1. Schemas may not be removed, only deprecated. Any changes to schemas must follow semantic versioning.
```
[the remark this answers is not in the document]
jischr: schemas cannot be deprecated explicity but superseded by newer versions

openid-sharedsignals-eventdef-specification-1_0.md
```diff
@@ -0,0 +1,277 @@
...
+- It enables a faster process to create, update and get approval for new event types. 
+- JSON schema, rather than spec texts, is a more appropriate format to describe event types. 
+- It also allows event types to be versioned independently thus reducing the friction between the SSF core specification and event type publication.
+- Allows more events to be incorporated easily in the standards space beyond the use cases and charters of CAEP, RISC etc.
+
+# JSON Schema Defintion
+
+The following section describes how to map a SSF event to a JSON Schema Document.
+
+As defined in Section 4.3 of {{JSON Schema}}, a JSON Schema document, also called a "schema", is a JSON document used to describe another JSON Document, known as an instance.
```
jischr, line 123:
```suggestion
As defined in Section 4.3 of {{JSON Schema}}, a JSON Schema document, also called a "schema", is a JSON document used to describe another JSON document, known as an instance.
```

openid-sharedsignals-eventdef-specification-1_0.md
```diff
@@ -0,0 +1,277 @@
...
+    },
+    "required": [ "initiating_entity" ]
+   }
+}
+~~~
+# Registry
+
+This section serves as a registry for the schemas of all registered SSF Event Types.
+
+| Event Type | Schema URI | Description |
```
jischr, line 244:
```suggestion
| Event Name | Schema URI | Description |
```

openid-sharedsignals-eventdef-specification-1_0.md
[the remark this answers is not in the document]
jischr: Fixed. Thanks for catching it.

openid-sharedsignals-eventdef-specification-1_0.md
jischr: Resolved this

openid-sharedsignals-eventdef-specification-1_0.md
jischr: Updated to Event Name

tulshi:
This is a great start, @jischr ! I would like to understand how a collection of events can be defined and identified as such here. For example, if someone wanted to implement "session security", then there may be many events that come under that collection (today named CAEP). Would there be a way to identify that in this scheme?

jischr:
> This is a great start, @jischr ! I would like to understand how a collection of events can be defined and identified as such here. For example, if someone wanted to implement "session security", then there may be many events that come under that collection (today named CAEP). Would there be a way to identify that in this scheme?

@tulshi  We could refer to this "Collection" in a new schema that described the whole events array.
It could then require two items of specified schemas - lets say 2 events total one of event type A and one of event type B. 
It could also require only one event : either event type A or event type B. 
Or any combination of schemas. 

I think the big takeaway from your comment is the distinction between event definitions and collections of event definitions.

Let me rough draft how that could look!

timcappalli:
>  For example, if someone wanted to implement "session security", then there may be many events that come under that collection (today named CAEP). Would there be a way to identify that in this scheme?

Wouldn't this just be addressed via an interop profile? Each event should be able to stand on its own in terms of understanding its usage. I don't think we should formally group events in this new model.

jischr:
decided not quite the right approach. still thing there is value in the idea. will see if i can come up with something better for future versions of ssf

Issue #211, opened 2024-10-08
Permanence of streams


tulshi:
Copying discussion from meeting on 9/24:

- [Tushar] the spec implicitly assumes that a stream is permanent once created. Should it support expirable streams?
- [Tushar] Possible solution: verification events on a cadence
- [Yair] A Rx (or Tx) can disable a stream
- [Yair] As long as PUSH endpoint is returning success, nothing is wrong. But if endpoint starts sending 500, 400 etc, then Tx can do something about it.
- [Yair] You cannot tell the difference between an Rx that has been sold, etc vs an Rx that is behaving well
- [Yair] We shouldn't make verification event requests mandatory, but can make it optional. Possibly include a time value that indicates how long between requests before Tx kills stream.
- [Shayne] I like the idea of an optional value in StreamConfiguration
- [Apoorva] We could also make it any call, not just verification
- [Shayne] Is that harder to implement?
- [Yair] Agree that it could be any endpoint - status endpoint would work
- [Shayne] That's nice because then you don't have to send unneccessary data

tulshi:
Action item recorded on 9/24: 
- [Tushar] Send PR to clarify that streams can optionally expire, and how to prevent that

keylimesoda:
+1 on Yair's comments.

It can be expensive for a transmitter to continue to transmit indefinitely to defunct receivers. It is potentially insecure to allow streams to exist indefinitely.

An optional "time remaining" or "expiration timedate" property on a Stream Configuration communicates a finite lifecycle to the Receiver.  The property could be updated by the Transmitter after successful verification event, or via PATCH request from receiver.

TusharR-google:
I'd like to propose that this be added to _vFinal_ instead of _vFuture_.

**Reasoning**
The main issue is security.

Permanent access to data (security events) for PUSH receivers is a potential security hole. We can contrast this with something like Oauth grants that are limited in both scope and duration.

The duration limit is missing, once the stream is opened by a PUSH Rx, the Tx is expected to continue sending events until the stream gets paused or deleted.

This also leads to asymmetry in the PUSH and PULL flows, where the PULL Rx is re-authed every time it pulls events, while the PUSH Rx may not communicate again for many years after creating the stream.

It would be great to have a way in the spec for Tx that are going to auto-expire streams (currently strongly preferred for the Google Tx implementation) to communicate this fact.

I'll send out a PR shortly.

**Cost Estimate vs ID3**
[...]

PR #222, opened 2024-12-03
Introduce inactivity_timeout

TusharR-google:
#211

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -99,7 +99,7 @@ normative:
     seriesinfo:
       DOI: 10.17487/RFC6749
       RFC: '6749'
-    target: https://datatracker.ietf.org/doc/html/rfc6749#appendix-A.14
+    target: https://tools.ietf.org/html/rfc6749#appendix-A.14
     title: The OAuth 2.0 Authorization Framework - "expires_in" Syntax
 
   OpenID.Core:
@@ -948,12 +948,12 @@ description
 
 stream_ttl
 
-> **Transmitter-Supplied**, OPTIONAL. The refreshable lifetime of the stream in seconds, after which the Transmitter MAY either pause or disable the stream if it has not received any Receiver-initiated communication (defined below) in that duration. The syntax is the same as that of {{EXPIRES_IN}}.
+> **Transmitter-Supplied**, OPTIONAL. The refreshable time-to-live (TTL) of the stream in seconds, after which the Transmitter MAY either pause or disable the stream if it has not received any Receiver-initiated communication (defined below) in that duration. The syntax is the same as that of {{EXPIRES_IN}}.
 >
-> For PUSH streams, the Transmitter MUST refresh the TTL whenever:
+> For streams created with the PUSH {{RFC8935}} delivery method, the Transmitter MUST refresh the TTL whenever:
 > * The Receiver calls any endpoint in the Event Stream Management API ({{management}}).
 >
-> For POLL streams, the Transmitter MUST refresh the TTL whenever:
+> For streams created with the POLL {{RFC8936}} delivery method, the Transmitter MUST refresh the TTL whenever:
 > * The Receiver polls the Transmitter for events.
 > * The Receiver calls any endpoint in the Event Stream Management API ({{management}}).
 >
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -1131,9 +1131,9 @@ description
   This is useful in multi-stream systems to identify the stream for human actors. The
   transmitter MAY truncate the string beyond an allowed max length.
 
-inactivity_ttl
+inactivity_timeout
 
-> **Transmitter-Supplied**, OPTIONAL. The refreshable time-to-live (TTL) of the stream in seconds. After the TTL passes with no eligible activity from the Receiver, as defined below, the Transmitter MAY either pause, disable, or delete the stream. The syntax is the same as that of `expires_in` from Section A.14 of {{RFC6749}}.
+> **Transmitter-Supplied**, OPTIONAL. The refreshable inactivity timeout of the stream in seconds. After the timeout duration passes with no eligible activity from the Receiver, as defined below, the Transmitter MAY either pause, disable, or delete the stream. The syntax is the same as that of `expires_in` from Section A.14 of {{RFC6749}}.
 >
 > The following constitutes eligible Receiver activity. If the Transmitter observes any of these activities from the Receiver, it MUST restart the inactivity timeout counter.
 >
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -1133,7 +1133,7 @@ description
 
 inactivity_ttl
 
-> **Transmitter-Supplied**, OPTIONAL. The refreshable time-to-live (TTL) of the stream in seconds, since the last activity from the Receiver as described below, was conducted. After the TTL passes with no eligible activity from the Receiver, the Transmitter MAY either pause, disable, or delete the stream. The syntax is the same as that of `expires_in` from Section A.14 of {{RFC6749}}.
+> **Transmitter-Supplied**, OPTIONAL. The refreshable time-to-live (TTL) of the stream in seconds. After the TTL passes with no eligible activity from the Receiver, as defined below, the Transmitter MAY either pause, disable, or delete the stream. The syntax is the same as that of `expires_in` from Section A.14 of {{RFC6749}}.
 >
 > The following constitutes eligible Receiver activity, upon observing which, the Transmitter MUST reset its count toward the TTL of the stream to zero.
 >
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -618,11 +618,11 @@ default_subjects
     to be transmitted. The Receiver MAY remove subjects added this way via the
     `remove_subject_endpoint`.
 
-push_stream_ttl
+stream_ttl
 
-> OPTIONAL. The lifetime of a PUSH stream in seconds, after which the Transmitter MAY either pause or disable the stream if it has not received any Receiver-initiated communication in that duration.
+> OPTIONAL. The lifetime of a stream in seconds, after which the Transmitter MAY either pause or disable the stream if it has not received any Receiver-initiated communication in that duration.
   If the Transmitter decides to update the stream, it MUST send a Stream Updated Event to the Receiver as described in {{status}}.
-  If the Receiver calls any endpoint in the Event Stream Management API ({{management}}), the Transmitter must refresh the TTL of that particular stream.
+  If the Receiver calls any endpoint in the Event Stream Management API ({{management}}), the Transmitter MUST refresh the TTL of that particular stream.
   The syntax is the same as that of {{EXPIRES_IN}}.
 
 TODO: consider adding a IANA Registry for metadata, similar to Section 7.1.1 of
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -1140,9 +1140,9 @@ description
   This is useful in multi-stream systems to identify the stream for human actors. The
   transmitter MAY truncate the string beyond an allowed max length.
 
-stream_ttl
+inactivity_ttl
 
-> **Transmitter-Supplied**, OPTIONAL. The refreshable time-to-live (TTL) of the stream in seconds, after which the Transmitter MAY either pause or disable the stream if it has not received any Receiver-initiated communication (defined below) in that duration. The syntax is the same as that of {{EXPIRES_IN}}.
+> **Transmitter-Supplied**, OPTIONAL. The refreshable time-to-live (TTL) of the stream in seconds, after which the Transmitter MAY either pause, disable, or delete the stream if it has not received any Receiver-initiated communication (defined below) in that duration. The syntax is the same as that of {{EXPIRES_IN}}.
 >
 > For streams created with the PUSH {{RFC8935}} delivery method, the Transmitter MUST refresh the TTL whenever:
 > * The Receiver calls any endpoint in the Event Stream Management API ({{management}}).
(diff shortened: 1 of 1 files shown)


appsdesh (dismissed):
Push and Pull are delivery mechanisms defined outside the SSF spec. We should not introduce new aspects specifically targeting a single delivery mechanism in the SSF spec

FragLegs:
> Push and Pull are delivery mechanisms defined outside the SSF spec. We should not introduce new aspects specifically targeting a single delivery mechanism in the SSF spec

I agree. Can we just make this `stream_ttl` so that it applies to any stream regardless of delivery method?

TusharR-google:
> Push and Pull are delivery mechanisms defined outside the SSF spec. We should not introduce new aspects specifically targeting a single delivery mechanism in the SSF spec

Thanks! The same concept can be applied to PULL streams too, let me reword this as simply stream_tll.

openid-sharedsignals-framework-1_0.md
```diff
@@ -601,6 +618,13 @@ default_subjects
     to be transmitted. The Receiver MAY remove subjects added this way via the
     `remove_subject_endpoint`.
 
+push_stream_ttl
+
+> OPTIONAL. The lifetime of a PUSH stream in seconds, after which the Transmitter MAY either pause or disable the stream if it has not received any Receiver-initiated communication in that duration.
+  If the Transmitter decides to update the stream, it MUST send a Stream Updated Event to the Receiver as described in {{status}}.
+  If the Receiver calls any endpoint in the Event Stream Management API ({{management}}), the Transmitter must refresh the TTL of that particular stream.
```
FragLegs, line 625:
```suggestion
  If the Receiver calls any endpoint in the Event Stream Management API ({{management}}), the Transmitter MUST refresh the TTL of that particular stream.
```

TusharR-google: Done

FragLegs: We might want to be explicit that contact from the receiver does not automatically move the stream out of a paused/disabled state caused by a TTL timeout.

openid-sharedsignals-framework-1_0.md
```diff
@@ -601,6 +618,13 @@ default_subjects
     to be transmitted. The Receiver MAY remove subjects added this way via the
     `remove_subject_endpoint`.
 
+stream_ttl
+
+> OPTIONAL. The lifetime of a stream in seconds, after which the Transmitter MAY either pause or disable the stream if it has not received any Receiver-initiated communication in that duration.
```
tulshi, line 623: "Receiver initiated" needs to be clarified, IMO. Does it differ for poll and push streams? Is the appropriate response code from the receiver to a push from a Tx sufficient?

openid-sharedsignals-framework-1_0.md
```diff
@@ -601,6 +618,13 @@ default_subjects
     to be transmitted. The Receiver MAY remove subjects added this way via the
     `remove_subject_endpoint`.
 
+stream_ttl
```
ptravers, line 621: can we make it `stream_ttl_seconds`? will save people working with the api a documentation lookup

openid-sharedsignals-framework-1_0.md
TusharR-google: Should we leave it up to the transmitter? It might complicate the spec if paused/disabled states were treated differently based on whether they were caused by a TTL timeout (Tx not allowed to automatically change the state), vs other causes like Rx requesting the pause (Tx allowed to do whatever).

openid-sharedsignals-framework-1_0.md
TusharR-google: Clarified the definition of "receiver-initiated" communication.

openid-sharedsignals-framework-1_0.md
TusharR-google: Most of the other names in the this/similar spec do not seem to use a suffix, regardless of units. E.g. min_verification_interval

So we could keep it as stream_ttl for consistency.

openid-sharedsignals-framework-1_0.md
```diff
@@ -85,6 +91,17 @@ normative:
     target: https://tools.ietf.org/html/rfc6749#section-4.4
     title: The OAuth 2.0 Authorization Framework - Client Credentials Grant
 
+  EXPIRES_IN:
```
tulshi, line 94: you don't need to describe an RFC. You can just add its number and the kramdown tool will do the rest.

openid-sharedsignals-framework-1_0.md
```diff
@@ -929,6 +946,19 @@ description
   This is useful in multi-stream systems to identify the stream for human actors. The
   transmitter MAY truncate the string beyond an allowed max length.
 
+stream_ttl
+
+> **Transmitter-Supplied**, OPTIONAL. The refreshable lifetime of the stream in seconds, after which the Transmitter MAY either pause or disable the stream if it has not received any Receiver-initiated communication (defined below) in that duration. The syntax is the same as that of {{EXPIRES_IN}}.
```
tulshi, line 951: You might want to expand the acronym "TTL" somewhere in the description. You say it precisely in the first sentence, but you just don't say what "TTL" stands for. It'll help clarify the name even though most folks will know what it is.

tulshi, line 951: Is the right behavior "pause or disable", or just "delete"?

TusharR-google: Pause / disable gives the Receiver an option to unpause / enable the stream, while delete means that the Receiver will have to recreate it.

I was imagining that some Transmitters may be more lenient where they simply pause the stream at first, then disable it after some time of being paused, and finally delete it if there's still no response (hence slowly escalating).

That said, either one works. Or we could even add the above paragraph as an example. What do you think?

TusharR-google: SG, done.

openid-sharedsignals-framework-1_0.md
TusharR-google: I wasn't sure how to make it point to a specific appendix, hence followed the example of CLIENTCRED above. Both CLIENTCRED and EXPIRES_IN are from the same RFC, so the details remain the same.

Would you know the right syntax to make it point to the appendix instead of the whole RFC (https://tools.ietf.org/html/rfc6749#appendix-A.14).

openid-sharedsignals-framework-1_0.md
FragLegs: Since it is MAY, you could say "...after which the Transmitter MAY either pause, disable, or delete the stream if it has not received..."

TusharR-google: Yeah we can keep the options open (implementation-dependent).

openid-sharedsignals-framework-1_0.md
```diff
@@ -1123,6 +1140,19 @@ description
...
+> **Transmitter-Supplied**, OPTIONAL. The refreshable time-to-live (TTL) of the stream in seconds, after which the Transmitter MAY either pause or disable the stream if it has not received any Receiver-initiated communication (defined below) in that duration. The syntax is the same as that of {{EXPIRES_IN}}.
+>
+> For streams created with the PUSH {{RFC8935}} delivery method, the Transmitter MUST refresh the TTL whenever:
+> * The Receiver calls any endpoint in the Event Stream Management API ({{management}}).
+>
+> For streams created with the POLL {{RFC8936}} delivery method, the Transmitter MUST refresh the TTL whenever:
+> * The Receiver polls the Transmitter for events.
+> * The Receiver calls any endpoint in the Event Stream Management API ({{management}}).
+>
+> If the Transmitter decides to update the stream, it MUST send a Stream Updated Event to the Receiver as described in {{status}}.
```
FragLegs, line 1154:
```suggestion
> If the Transmitter decides to pause or disable the stream, it MUST send a Stream Updated Event to the Receiver as described in {{status}}.
```
It took me a second to understand why this sentence was here. Specifying pause or disable connects it to the language above.

openid-sharedsignals-framework-1_0.md
```diff
@@ -1123,6 +1140,19 @@ description
   This is useful in multi-stream systems to identify the stream for human actors. The
   transmitter MAY truncate the string beyond an allowed max length.
 
+inactivity_ttl
+
+> **Transmitter-Supplied**, OPTIONAL. The refreshable time-to-live (TTL) of the stream in seconds, after which the Transmitter MAY either pause, disable, or delete the stream if it has not received any Receiver-initiated communication (defined below) in that duration. The syntax is the same as that of {{EXPIRES_IN}}.
+>
+> For streams created with the PUSH {{RFC8935}} delivery method, the Transmitter MUST refresh the TTL whenever:
+> * The Receiver calls any endpoint in the Event Stream Management API ({{management}}).
```
tulshi, line 1148: We probably need to specify here and below that "calls to the Event Stream Management API that reference the stream"

openid-sharedsignals-framework-1_0.md
```diff
@@ -1123,6 +1140,19 @@ description
   This is useful in multi-stream systems to identify the stream for human actors. The
   transmitter MAY truncate the string beyond an allowed max length.
 
+inactivity_ttl
+
+> **Transmitter-Supplied**, OPTIONAL. The refreshable time-to-live (TTL) of the stream in seconds, after which the Transmitter MAY either pause, disable, or delete the stream if it has not received any Receiver-initiated communication (defined below) in that duration. The syntax is the same as that of {{EXPIRES_IN}}.
+>
+> For streams created with the PUSH {{RFC8935}} delivery method, the Transmitter MUST refresh the TTL whenever:
```
tulshi, line 1147: "refresh the TTL" could be ambiguous. Can we be more precise? we could alter the description in line 1145 to read: "the refreshable time-to-live (TTL) of the stream in seconds, since the last activity from the Receiver as described below, was conducted. After the TTL, the Transmitter MAY..."

This sentence can then read: "Receiver activity can be any of:"

openid-sharedsignals-framework-1_0.md
tulshi: I guess you aren't really referring to "CLIENTCRED", right? The Expires_in definition in OAuth is pretty small, so one way you can do this is by referring to it as "Section A.14 of  RFC6749 {{RFC6749}}".

openid-sharedsignals-framework-1_0.md
```diff
@@ -1123,6 +1140,19 @@ description
   This is useful in multi-stream systems to identify the stream for human actors. The
   transmitter MAY truncate the string beyond an allowed max length.
 
+inactivity_ttl
+
+> **Transmitter-Supplied**, OPTIONAL. The refreshable time-to-live (TTL) of the stream in seconds, after which the Transmitter MAY either pause, disable, or delete the stream if it has not received any Receiver-initiated communication (defined below) in that duration. The syntax is the same as that of {{EXPIRES_IN}}.
```
tulshi, line 1145: see suggestion above.

openid-sharedsignals-framework-1_0.md
TusharR-google: Done

openid-sharedsignals-framework-1_0.md
TusharR-google: Done

tulshi (approved):

openid-sharedsignals-framework-1_0.md
TusharR-google: Done in a later commit as suggested

TusharR-google:
> Push and Pull are delivery mechanisms defined outside the SSF spec. We should not introduce new aspects specifically targeting a single delivery mechanism in the SSF spec

#

FragLegs (approved):

openid-sharedsignals-framework-1_0.md
```diff
@@ -1123,6 +1131,23 @@ description
   This is useful in multi-stream systems to identify the stream for human actors. The
   transmitter MAY truncate the string beyond an allowed max length.
 
+inactivity_ttl
+
+> **Transmitter-Supplied**, OPTIONAL. The refreshable time-to-live (TTL) of the stream in seconds, since the last activity from the Receiver as described below, was conducted. After the TTL passes with no eligible activity from the Receiver, the Transmitter MAY either pause, disable, or delete the stream. The syntax is the same as that of `expires_in` from Section A.14 of {{RFC6749}}.
```
FragLegs, line 1136:
```suggestion
> **Transmitter-Supplied**, OPTIONAL. The refreshable time-to-live (TTL) of the stream in seconds. After the TTL passes with no eligible activity from the Receiver, as defined below, the Transmitter MAY either pause, disable, or delete the stream. The syntax is the same as that of `expires_in` from Section A.14 of {{RFC6749}}.
```
Nit suggestion for clarity

tulshi:
> Push and Pull are delivery mechanisms defined outside the SSF spec. We should not introduce new aspects specifically targeting a single delivery mechanism in the SSF spec

Since the comments have been resolved in the PR, I'm going to "dismiss" this review. @appsdesh please comment if you think this still needs addressing.

openid-sharedsignals-framework-1_0.md
```diff
@@ -1123,6 +1131,23 @@ description
...
+>
+> The following constitutes eligible Receiver activity, upon observing which, the Transmitter MUST reset its count toward the TTL of the stream to zero.
```
FragLegs, line 1138:
```suggestion
> The following constitutes eligible Receiver activity. If the Transmitter observes any of these activities from the Receiver, it MUST restart the inactivity timeout counter.
```
Nit suggestion for clarity

FragLegs (approved):

tulshi (approved):

FragLegs (approved):
This looks great! Thanks for all the diligence getting it tuned up so well.

iamseanodentity (approved):

openid-sharedsignals-framework-1_0.md
```diff
@@ -1123,6 +1131,23 @@ description
   This is useful in multi-stream systems to identify the stream for human actors. The
   transmitter MAY truncate the string beyond an allowed max length.
 
+inactivity_timeout
+
+> **Transmitter-Supplied**, OPTIONAL. The refreshable inactivity timeout of the stream in seconds. After the timeout duration passes with no eligible activity from the Receiver, as defined below, the Transmitter MAY either pause, disable, or delete the stream. The syntax is the same as that of `expires_in` from Section A.14 of {{RFC6749}}.
```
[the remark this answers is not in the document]
TusharR-google: For PUSH streams that's not enough, at least from a security/privacy perspective - receiver endpoints are more likely to be open, the only way to verify that the receiver exists and still has the right auth/access is for it to call the transmitter endpoints (oauth-protected in most cases).

For PULL, polling is sufficient.

PR #206, opened 2024-09-20
IP Addresses as a format

appsdesh:
IP Addresses as a format

Closes #201

PR #237, opened 2025-02-25
IP Addresses as a format

appsdesh:
IP Addresses as a format

Closes https://github.com/openid/sharedsignals/issues/201

continuing from https://github.com/openid/sharedsignals/pull/206


openid-caep-1_0.md
```diff
@@ -735,9 +735,6 @@ The `event_timestamp` in this event type specifies the time at which the session
 ### Event Specific Claims {#session-established-event-specific-claims}
 The following optional claims MAY be included in the Session Established event:
 
-ips
-: The array of IP addresses of the user as observed by the Transmitter. The value MUST be in the format of an array of strings, each one of which represents the RFC 4001 {{RFC4001}} string representation of an IP address. (**NOTE**, this can be different from the one observed by the Receiver for the same user because of network translation)
```
tulshi, line 739: Please check if we need the RFC4001 reference in this spec, now that this text has been removed. If it is not needed, please remove that reference.

timcappalli (dismissed):

tulshi (approved):
I prefer ips, but ip-addresses is OK too.

PR #243, opened 2025-03-31
Fixes issue #208 - Poll endpoint should require authorization

### LINKED ISSUE #208: Poll endpoint should require authorization
The second recommendation from the final security audit:

As we note in Section 2.6, poll endpoint URLs
are not required to be secret, i.e., SETs could be requested by any party. For use cases requiring
confidentiality of SETs, we recommend mandating authorization at the poll endpoint.

jischr:
Add text to section [7.1.1](https://openid.github.io/sharedsignals/openid-sharedsignals-framework-1_0.html#section-7.1.1) to state that authorization should be used for the Poll Endpoint

Fixes #208

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -78,7 +78,11 @@ contributor:
         name: James Slocum
         org: Beyond Identity
         email: james.slocum@beyondidentity.com
-
+      -
+        ins: J. Schreiber
+        name: Jen Schreiber
+        org: Workday
+        email: jennifer.winer@workday.com
 normative:
   CLIENTCRED:
     author:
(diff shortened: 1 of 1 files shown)


appsdesh (commented):
I suggest replacing `SHOULD` with a `MAY`

That's the stance we take in the interop profile [here](https://openid.net/specs/openid-caep-interoperability-profile-1_0-ID1.html#section-2.7.2-2.2.1)

jischr:
@appsdesh What about replaces with its RECOMMENDED

FragLegs (approved):

iamseanodentity (approved):
Love lines 81 - 85 :)
Ship it

PR #244, opened 2025-03-31
Fixes #221 - Add event_metadata to CAEP and RISC

### LINKED ISSUE #221: Addition of Event data field inside events
The intent is to be able to offer guidelines for all events in order to allow transmitters to provide more metadata that can be specific to the event and or action(s) in the event.

example:
```
  "events": {
    "https://schemas.openid.net/secevent/caep/event-type/session-revoked": {
      "event_timestamp": 1615304991643,
      "event_data": {
        "something":"foo",
        "blah":"bar",
        "blah2":"example2"
      }
    }
  }
```

jischr:
Added event_metadata to CAEP and RISC for the transmitter to provide additional information about the event itself.

Fixes #221

### DIFF
--- a/openid-caep-1_0.md
+++ b/openid-caep-1_0.md
@@ -174,10 +174,9 @@ event_timestamp
   from 1970-01-01T0:0:0Z as measured in UTC until the date/time.
 
 event_metadata
-: OPTIONAL, JSON object: metadata specific to the event and or actions(s) in
- the event provided by the transmitter. The object MUST contain one or more
- key/value pairs. This claim SHOULD not be used to express information about
- the subject.
+: OPTIONAL, JSON object: metadata specific to the event provided by the
+Transmitter. The object MUST contain one or more key/value pairs. This claim
+SHOULD be used to express information about the event, not the subject.
 
 initiating_entity
 : OPTIONAL, JSON string: describes the entity that invoked the event.
--- a/openid-caep-1_0.md
+++ b/openid-caep-1_0.md
@@ -173,6 +173,12 @@ event_timestamp
   occurred. Its value is a JSON number representing the number of seconds
   from 1970-01-01T0:0:0Z as measured in UTC until the date/time.
 
+event_metadata
+: OPTIONAL, JSON object: metadata specific to the event and or actions(s) in
+ the event provided by the transmitter. The object MUST contain one or more
+ key/value pairs. This claim SHOULD not be used to express information about
+ the subject.
+
 initiating_entity
 : OPTIONAL, JSON string: describes the entity that invoked the event.
 : This MUST be one of the following strings:
--- a/openid-risc-profile-specification-1_0.xml
+++ b/openid-risc-profile-specification-1_0.xml
@@ -94,6 +94,16 @@
       <t>The base URI for RISC event types is:<vspace />
         https://schemas.openid.net/secevent/risc/event-type/</t>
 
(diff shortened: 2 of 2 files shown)


appsdesh (changes requested):
- what does it mean by action?
- how would we prevent bloating?
- Why are we singling this out  "This claim SHOULD not be used to express information about
  the subject." needs more concrete guidance

jischr:
@appsdesh 

> * what does it mean by action? 
- that wording was taken from the issue. happy to remove or replace if it doesnt suit.

> * how would we prevent bloating? 
- I don't believe we need to prevent bloating as SETs do not have restrictions in place for event claims. if necessary, we could put restrictions on string length.

> * Why are we singling this out  "This claim SHOULD not be used to express information about
>   the subject." needs more concrete guidance
- We need to differentiate the this claim for other claims that could have been added to the individual event schemas (that describes the subject). this is metadata. Guidance could be added to prevent this claim from becoming out of hand.

openid-caep-1_0.md
```diff
@@ -173,6 +173,12 @@ event_timestamp
   occurred. Its value is a JSON number representing the number of seconds
   from 1970-01-01T0:0:0Z as measured in UTC until the date/time.
 
+event_metadata
+: OPTIONAL, JSON object: metadata specific to the event and or actions(s) in
+ the event provided by the transmitter. The object MUST contain one or more
+ key/value pairs. This claim SHOULD not be used to express information about
+ the subject.
```
FragLegs, line 180:
```suggestion
 the event rather than the subject.
```

iamseanodentity:
"action" here is needed, but we do not need to call it out in the description. In the event_metadata field it is setting the foundation to expand upon profile events that might be mandatory via a contract between the transmitter and receiver. 

We need guidelines for this and a use case with event_metadata details.
- Don't overload this claim (size considerations of the JWT)
- Don't add in subject based information and keep it siloed to the event itself (i.e. correlation_id, enriched meta data info)
- Why and When to use it
-

openid-caep-1_0.md
```diff
@@ -173,6 +173,11 @@ event_timestamp
   occurred. Its value is a JSON number representing the number of seconds
   from 1970-01-01T0:0:0Z as measured in UTC until the date/time.
 
+event_metadata
```
appsdesh, line 176:
```suggestion
metadata
```

openid-risc-profile-specification-1_0.xml
```diff
@@ -94,6 +94,16 @@
       <t>The base URI for RISC event types is:<vspace />
         https://schemas.openid.net/secevent/risc/event-type/</t>
 
+      <section anchor="optional-event claims" title="Optional Event Claims">
+        <t>The following claims are optional unless otherwise specified in the event type definition.
+
+        <list style="symbols">
+          <t>event_metadata - optional, JSON object,metadata specific to the event provided by the
```
appsdesh, line 101:
```suggestion
          <t>metadata - optional, JSON object,metadata specific to the event provided by the
```

appsdesh (approved):
New changes look good. Suggesting a nit to remove redundant `event` from the key

jischr:
@iamseanodentity do you can an example of this you can supply?

iamseanodentity:
> @iamseanodentity do you can an example of this you can supply?

I will get to this tonight. Would you like me to add inline or just here and you can handle the copy?

jischr:
decided against this 4/15

PR #246, opened 2025-04-01
Adding some language for recommended stream configuration audience validation

### LINKED ISSUE #207: Receivers should validate aud value in StreamConfiguration response
The first recommendation from the final security audit:

While Receivers are mandated to validate
the audience value in SETs (due to [RFC7519, Section 4.1.3]), they are currently not required to
validate the audience value in stream configurations returned by a Transmitter, e.g., in a stream
creation response. Our Receiver model respects this and hence mostly ignores streams’ audience
values. For SET validation, our Receiver model instead compares the SET’s audience value against
an expected value based on the access token used by the Receiver when requesting creation of
the stream (since this is where the Transmitter is required to derive an audience value from the
Receiver’s authorization, see [15, Section 7]).
However, it is likely that implementers use the stream’s audience value to validate SETs against,
hence, we recommend to mandate Receivers-side validation of stream audience values.

jischr:
This is an attempt at some language for recommending the validation of the audience in the stream configuration.

It seems quite vague.... As @FragLegs mentions in the issue, it could be good to be more concrete on what this audience value should be.

Fixes #207

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -1229,14 +1229,20 @@ Errors are signaled with HTTP status codes as follows:
 {: title="Create Stream Errors" #tablecreatestream}
 
 
+##### Validating a Stream Creation Response
+
+* `aud`: It is RECOMMENDED that the Receiver validates the `aud` in the Create Stream Response.
+In most cases, a Transmitter and Receiver will agree upon the audience value out of band. 
+Regardless of how the Receiver obtains the audience, it SHOULD ensure that it matches the
+response.
+
 #### Reading a Stream’s Configuration {#reading-a-streams-configuration}
 An Event Receiver gets the current configuration of a stream by making an HTTP
 GET request to the Configuration Endpoint. On receiving a valid request, the
 Event Transmitter responds with a "200 OK" response containing a [JSON][RFC7159]
 representation of the stream’s configuration in the body.  The Receiver
 MUST check the response and confirm that the `iss` value matches the Issuer from
-which it received the Transmitter Configuration data. It is RECOMMENDED that the
-Receiver validates the `aud` value matches what they expect.
+which it received the Transmitter Configuration data.
 
 The GET request MAY include the "stream_id" as a query parameter in order to
 identify the correct Event Stream. If the "stream_id" parameter is missing,
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -1236,7 +1236,7 @@ Event Transmitter responds with a "200 OK" response containing a [JSON][RFC7159]
 representation of the stream’s configuration in the body.  The Receiver
 MUST check the response and confirm that the `iss` value matches the Issuer from
 which it received the Transmitter Configuration data. It is RECOMMENDED that the
-Receiver validates the `aud` matches the expected value.
+Receiver validates the `aud` value matches what they expect.
 
 The GET request MAY include the "stream_id" as a query parameter in order to
 identify the correct Event Stream. If the "stream_id" parameter is missing,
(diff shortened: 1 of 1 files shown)


openid-sharedsignals-framework-1_0.md
```diff
@@ -1229,6 +1229,13 @@ Errors are signaled with HTTP status codes as follows:
 {: title="Create Stream Errors" #tablecreatestream}
 
 
+##### Validating a Stream Creation Response
+
+* `aud`: It is RECOMMENDED that the Receiver validates the `aud` in the Create Stream Response.
```
tulshi, line 1234: I think you need to rephrase the sentence to use the word "SHOULD" (i.e., The Receiver SHOULD validate the 'aud' value in the Create Stream Response.

openid-sharedsignals-framework-1_0.md
```diff
@@ -1229,6 +1229,13 @@ Errors are signaled with HTTP status codes as follows:
...
+In most cases, a Transmitter and Receiver will agree upon the audience value out of band.
```
tulshi, line 1235: Use normative language here too. Drop the "In most cases", and say "A Transmitter and Receiver MAY agree upon the audience value out of band."

openid-sharedsignals-framework-1_0.md
```diff
@@ -1229,6 +1229,13 @@ Errors are signaled with HTTP status codes as follows:
 {: title="Create Stream Errors" #tablecreatestream}
 
 
+##### Validating a Stream Creation Response
+
+* `aud`: It is RECOMMENDED that the Receiver validates the `aud` in the Create Stream Response.
+In most cases, a Transmitter and Receiver will agree upon the audience value out of band. 
+Regardless of how the Receiver obtains the audience, it SHOULD ensure that it matches the
```
tulshi, line 1236: "obtains" doesn't sound right here. How about "Regardless of how the audience value is agreed upon"?

openid-sharedsignals-framework-1_0.md
jischr:
```suggestion
* `aud`: the Receiver SHOULD validate the `aud` in the Create Stream Response.
```

openid-sharedsignals-framework-1_0.md
```diff
@@ -1229,6 +1229,12 @@ Errors are signaled with HTTP status codes as follows:
 {: title="Create Stream Errors" #tablecreatestream}
 
 
+##### Validating a Stream Creation Response
+
+* `aud`: the Receiver SHOULD validate the `aud` in the Create Stream Response.
+A Transmitter and Receiver MAY agree upon the audience value out of band.
```
tulshi, line 1235: The formatting doesn't look right. Have you tried making it locally and checking the output?

jischr: Looks ok? 
Screenshot 2025-04-24 at 10 09 35 AM

openid-sharedsignals-framework-1_0.md
```diff
@@ -1229,6 +1229,12 @@ Errors are signaled with HTTP status codes as follows:
...
+Regardless of how the audience value is agreed upon, the Receiver SHOULD ensure that it matches what is expected.
```
tulshi, line 1236: One nit: Can we say "matches what it expects", rather than "matches what is expected" (clarifies who sets the expectation).

jischr, line 1236:
```suggestion
Regardless of how the audience value is agreed upon, the Receiver SHOULD ensure that it matches what it expects.
```

tulshi (approved):

iamseanodentity (approved):

FragLegs (approved):

PR #247, opened 2025-04-01
Clarify Stream Creation default delivery method behavior

### LINKED ISSUE #149: Stream creation lanaguage imply that transmiter must suport the poll delivery method if delivery method is not specified
Section 7.1.1.1 (stream creation) have the following language: "_If the request does not contain the delivery property, then the Transmitter MUST assume that the method is "urn:ietf:rfc:8936" (poll). The Transmitter MUST include a delivery property in the response with this method property and an endpoint_url property._"

The language seem to imply that the transmitter must support the poll method, which shouldn't be the case as supporting the poll method is optional.

The language should be changed that a 400 response should be returned if the poll method is not supported when delivery is not specified in the request.

jischr:
Fixes #149 

This PR should be non-normative (is it?!)  i've tried to better describe the default behavior when a transmitter does not support the defaulted stream delivery method.

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -1154,10 +1154,12 @@ Configuration ({{stream-config}}) object:
 * `description`
 
 If the request does not contain the `delivery` property, then the Transmitter
-MUST assume that the `method` is "urn:ietf:rfc:8936" (poll). The
-Transmitter MUST include a `delivery` property in the response with this
-`method` property and an `endpoint_url` property. Note that in the case of the poll
-method, the `endpoint_url` value is supplied by the Transmitter.
+MUST assume that the `method` is "urn:ietf:rfc:8936" (poll). If the Transmitter supports
+Poll-Based Delivery, the Transmitter MUST include a `delivery` property in the response with this
+`method` property and an `endpoint_url` property. If the Transmitter does not support
+the delivery method, it MAY respond with 4XX HTTP Status Code such as "400 Bad Rquest."
+
+Note that in the case of the poll method, the `endpoint_url` value is supplied by the Transmitter.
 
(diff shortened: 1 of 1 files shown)


FragLegs:
I believe that @atultulshi said we should not include this in the next release because it is a backwards-incompatible change: #149

openid-sharedsignals-framework-1_0.md
```diff
@@ -1154,10 +1154,12 @@ Configuration ({{stream-config}}) object:
...
 
 If the request does not contain the `delivery` property, then the Transmitter
-MUST assume that the `method` is "urn:ietf:rfc:8936" (poll). The
-Transmitter MUST include a `delivery` property in the response with this
-`method` property and an `endpoint_url` property. Note that in the case of the poll
-method, the `endpoint_url` value is supplied by the Transmitter.
+MUST assume that the `method` is "urn:ietf:rfc:8936" (poll). If the Transmitter supports
+Poll-Based Delivery, the Transmitter MUST include a `delivery` property in the response with this
+`method` property and an `endpoint_url` property. If the Transmitter does not support
+the delivery method, it can respond with 4XX HTTP Status Code such as "400 Bad Rquest."
```
FragLegs, line 1160:
```suggestion
the delivery method, it MAY respond with 4XX HTTP Status Code such as "400 Bad Rquest."
```

jischr, line 1160:
```suggestion
the delivery method, it MAY respond with HTTP Status Code "400 Bad Request."
```

FragLegs (approved):

FragLegs:
For anyone reading the comments - we adjusted the language so that it is fully backwards compatible and just gives a suggestion for how to respond if the Tx does not support polling.

PR #248, opened 2025-04-02
Update error table

appsdesh:
A follow-up from https://github.com/openid/sharedsignals/pull/247

@jischr please take a look

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -1232,7 +1232,7 @@ Errors are signaled with HTTP status codes as follows:
 
 | Code | Description |
 |------|-------------|
-| 400  | if the request cannot be parsed, delivery method not supported |
+| 400  | if the request cannot be parsed, or if the delivery method is not supported |
 | 401  | if authorization failed or it is missing |
 | 403  | if the Event Receiver is not allowed to create a stream |
 | 409  | if the Transmitter does not support multiple streams per Receiver |
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -1232,7 +1232,7 @@ Errors are signaled with HTTP status codes as follows:
 
 | Code | Description |
 |------|-------------|
-| 400  | if the request cannot be parsed |
+| 400  | if the request cannot be parsed, delivery method not supported |
 | 401  | if authorization failed or it is missing |
 | 403  | if the Event Receiver is not allowed to create a stream |
 | 409  | if the Transmitter does not support multiple streams per Receiver |


jischr (approved):
Looks good. Thanks for adding this to the table!

openid-sharedsignals-framework-1_0.md
```diff
@@ -1232,7 +1232,7 @@ Errors are signaled with HTTP status codes as follows:
 
 | Code | Description |
 |------|-------------|
-| 400  | if the request cannot be parsed |
+| 400  | if the request cannot be parsed, delivery method not supported |
```
tulshi, line 1235: This is a bit ambiguous. Did you mean "if the request cannot be parsed, or the delivery method is not supported"?

jischr, line 1235:
```suggestion
| 400  | if the request cannot be parsed, or if the delivery method is not supported |
```

tulshi:
> Returning a 400 if the delivery method is not supported. I looked through the 4xx error codes. Can we suggest providing a reason phrase with the code to delineate why it's a bad request? Thoughts @jischr

So this is a bug in the current spec?

gffletch:
Would using HTTP status code 406 be better or applicable?

tulshi:
The current language says "errors are reported...", which makes the current behavior the equivalent of a "MUST", changing this will break backward compatibility. There is no evidence right now that this is an implementation blocker, so we can review this after v1Final.

tulshi:
> > The current language says "errors are reported...", which makes the current behavior the equivalent of a "MUST", changing this will break backward compatibility. There is no evidence right now that this is an implementation blocker, so we can review this after v1Final.
> 
> Not sure I understand your comment @atultulshi , this PR was to bring alignment of the text which describes the error and the table which lists these scenarios.

Good point. I think in the discussion around this PR we confused the topics of requesting verification events when the stream was paused or disabled. We decided to keep that one out of the scope of v1Final. I won't be there, but you can discuss this PR separately in the meeting on 4/29.

FragLegs:
@appsdesh The text says that the Transmitter MAY send the 400 code if the delivery method is not available. But the intro to the table says "Errors are signaled" which should be understood more like a MUST. Adding the delivery unavailability to the table would break backwards compatibility. 

The solution might be to go through the spec and reword the error tables so that they say MAY in the intro. But that seems contentious enough to push off until after v1.

appsdesh:
> @appsdesh The text says that the Transmitter MAY send the 400 code if the delivery method is not available. But the intro to the table says "Errors are signaled" which should be understood more like a MUST. Adding the delivery unavailability to the table would break backwards compatibility.
> 
> The solution might be to go through the spec and reword the error tables so that they say MAY in the intro. But that seems contentious enough to push off until after v1.

We are creating two sources of truth for the error codes, one in the table and another buried in the description. I understand MAY vs MUST, but we are likely making it difficult for the implementers to see a list of all expected/suggested errors. We should align on this holistically for v1 and not punt it.

Issue #255, opened 2025-05-12
SSF Events not just informational


gffletch:
Whenever I bring up the use of SSF for managing session state or changes to deployed systems, I quite often get a response that SSF Events are "just informational". In my conversations with a number of people this isn't strictly true. Instead, the SSF specifications are silent on the behavioral rules in play for any given deployment when an event is received.

I'm wondering if makes sense to clearly call this out in the specification as a non-normative information statement. Something like...

The Shared Signals Framework does not define explicit processing behavior for receipt of events specifically to allow for each deployment to define the behaviors that make sense for that environment. These behaviors can range from treating the event as informational input to additional processing, to mandatory enforcement of the specified state change (e.g. session revoke).

timcappalli:
It definitely isn't in the spec, but we had been talking about that being a business decision/agreement based on the relationship between a transmitter and receiver. 

That could also be in a consortium-like relationship (e.g. in this network, if you get session revoked, you're expecting to send the user back to the IdP).

But this was always just verbal/positioning and nothing normative outside of what Karl highlighted from SET.

tulshi:
I understand the last paragraph in Section 1 of RFC8417 to be about that "an SET cannot be assumed to be commands or requests". I would have added the word "automatically" there. In section 3, I assume the word "Semantics" to include the kind of command semantics we are talking about here, even though the example provided is about the format.

In conclusion, in Section 1, or Section 3 (or anywhere else in RFC8417), I don't see language that prohibits the use of SETs as commands, so although they are defined to be informational, using them for commands is not ruled out.

mcguinness:
Command is flipping "may interpret for their own purposes" in the last paragraph in Section 1 of RFC8417 to "MUST interpret as X".  SETs are defined as informational as to not prescribe how the consumer MUST interpret the event.  This decoupling was why they were modeled as events.  Its just an observable.

dhs-BI:
See this [thread](https://github.com/openid/ipsie/issues/66) from the IPSIE WG as well.

My interest is in using profiles to define how the recipient of an event takes action on that event.  In other words, we can further constrain the existing profiles and assign an action to event receipt if, and only if, both parties are agreeing to be bound by the requirements of the profile.  

IPSIE will tackle this issue sometime down the road, most likely no earlier than 2026 given the work that remains to be completed between now and our proposed interop event in December 2025.

PR #267, opened 2025-05-29
Generic Language for prescriptive SETs - Issue 255

iamseanodentity:
In reference to [Issue 255](https://github.com/openid/sharedsignals/issues/255) marked for v1 Final per the discussion last week in the SSWG Call. 

I added a sub section under Section 4 called "SSF Prescriptive SETs" and an example figure of what a prescriptive SET may look like under Section 5.

Open to feedback on the example. I took the IPSIE action out as defined [here](https://github.com/openid/ipsie/issues/66) and replaced it with a duplicative `session-revoked` event to convey that the transmitter is expecting the action to be taken is to revoke a session even though the SET is referencing the same event. I am all for using an IPSIE profile in there to convey cross standard usage of SSF but wanted to get some feedback from the reviewers. 

I thought Section 4 was a decent place to call out that the Events in a SET can be treated as prescriptive. For those following along I will add the text here for brevity. I tried to keep it clean and concise and added a new section called **SSF Prescriptive SETs**. See below:
```The Shared Signals Framework allows each deployment or integration to define its own event processing behaviors, ranging from informational input to additional processing needed, to mandatory enforcement (e.g., session revoke). See {{Referenced Example I created in Section 5}}.```

### DIFF
--- a/openid-sharedsignals-framework-1_0.md
+++ b/openid-sharedsignals-framework-1_0.md
@@ -484,7 +484,7 @@ on this subject. The Shared Signals Framework requires further restrictions:
 The signature key can be obtained through "jwks_uri", see {{discovery}}.
 
 ### SSF Prescriptive SETs {#prescriptive-sets}
-The Shared Signals Framework allows each deployment or integration to define its own event processing behaviors, ranging from informational input to additional processing needed, to mandatory enforcement (e.g., session revoke). See {{caep-event-prescriptive-example}}.
+The Shared Signals Framework allows each deployment or integration to define its own event processing behaviors, ranging from informational input to additional processing needed, to mandatory enforcement.
 
 ### The "iss" Claim {#iss-claim}
 The "iss" claim MUST match the "iss" value in the Stream Configuration data for the stream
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: openid-sharedsignals-framework-1_0.md [hunk 1]
```diff
@@ -694,6 +697,27 @@ The following are hypothetical examples of SETs that conform to the Shared Signa
...
+  "txn": 8675309,
+  "aud": "636C69656E745F6964",
+  "sub_id": {
+    "format": "phone_number",
+    "phone_number": "+1 813 867 5309"
+  },
+  "events": {
+    "https://schemas.openid.net/secevent/caep/event-type/session-revoked": {}
+  },
+  "actions": {
```
mcguinness, line 714: If a profile is just changing processing semantics of the secevent but not changing the event itself, why is there a need to extend the event with an `actions` claim?

mcguinness: I strongly disagree that we should be extending the secevent schema to support mandatory actions uris.  I thought the goal of #255 was just to allow profiles to define additional processing semantics of an event not change the event itself to describe actions.  I made similar comments on the referenced issued.

FragLegs: I agree with Karl here. The `actions` field was a proposed normative way that we could make SSF prescriptive. I think the intent of this PR, at this stage of v1, is just to add some text that says "If you want to use Shared Signals in a prescriptive way, the SSF police won't come after you about it." I think your paragraph above does that well and this example just confuses things.

openid-sharedsignals-framework-1_0.md
```diff
@@ -483,6 +483,9 @@ on this subject. The Shared Signals Framework requires further restrictions:
 ### Signature Key Resolution {#signature-key-resolution}
 The signature key can be obtained through "jwks_uri", see {{discovery}}.
 
+### SSF Prescriptive SETs {#prescriptive-sets}
+The Shared Signals Framework allows each deployment or integration to define its own event processing behaviors, ranging from informational input to additional processing needed, to mandatory enforcement (e.g., session revoke). See {{caep-event-prescriptive-example}}.
```
FragLegs, line 487:
```suggestion
The Shared Signals Framework allows each deployment or integration to define its own event processing behaviors, ranging from informational input to additional processing needed, to mandatory enforcement.
```
Calling out `session-revoked` here is confusing because then I am left wondering if that event is always meant to be prescriptive.

tulshi (approved):

openid-sharedsignals-framework-1_0.md
iamseanodentity: Done and updated

PR #277, opened 2025-06-23
Changed examples to show txn value is a string.

tulshi:
fixes issue: 273


FragLegs (changes requested):
There are also 13 instances in the CAEP spec

FragLegs (approved):
Never mind. Those are being addressed in https://github.com/openid/sharedsignals/pull/279

Issue #274, opened 2025-06-22
editorial issues in the SSF spec under public review


TakahikoKawasaki:
### Abstract

```diff
- A profile for Security Events Tokens [RFC8417]
+ A profile for Security Event Token [RFC8417]
```

```diff
- Push-Based SET Token Delivery Using HTTP [RFC8935]
- Poll-Based SET Token Delivery Using HTTP [RFC8936]
+ Push-Based Security Event Token (SET) Delivery Using HTTP [RFC8935]
+ Poll-Based Security Event Token (SET) Delivery Using HTTP [RFC8936]
```

### 3.1. Subject Members

```diff
- A top-level claimnamed
+ A top-level claim named
```

### 3.3. Complex Subject Members

```diff
  "sub_id": {
    "format": "complex",
-   "user" : {
+   "user": {
      "format": "email",
      "email": "bar@example.com"
    },
-   "tenant" : {
+   "tenant": {
      "format": "iss_sub",
-     "iss" : "https://example.com/idp1",
-     "sub" : "1234"
+     "iss": "https://example.com/idp1",
+     "sub": "1234"
    }
  }
```

### 3.5.1. JWT ID Subject Identifier Format

```diff
  {
-     "format": "jwt_id",
-     "iss": "https://idp.example.com/123456789/",
-     "jti": "B70BA622-9515-4353-A866-823539EECBC8"
+   "format": "jwt_id",
+   "iss": "https://idp.example.com/123456789/",
+   "jti": "B70BA622-9515-4353-A866-823539EECBC8"
  }
```

### 3.5.2. SAML Assertion ID Subject Identifier Format

```diff
  {
-     "format": "saml_assertion_id",
-     "issuer": "https://idp.example.com/123456789/",
-     "assertion_id": "_8e8dc5f69a98cc4c1ff3427e5ce34606fd672f91e6"
+   "format": "saml_assertion_id",
+   "issuer": "https://idp.example.com/123456789/",
+   "assertion_id": "_8e8dc5f69a98cc4c1ff3427e5ce34606fd672f91e6"
  }
- 
```

### 3.5.3. IP Addresses Subject Identifier Format

```diff
  {
-     "format": "ip-addresses",
-     "ip-addresses": ["10.29.37.75", "2001:0db8:0000:0000:0000:8a2e:0370:7334"]
+   "format": "ip-addresses",
[...]

PR #278, opened 2025-06-23
Fixes issue #274 - editorial issues in the SSF spec under public review

TakahikoKawasaki:


jischr (approved):
pending comment by shayne

openid-sharedsignals-framework-1_0.md
```diff
@@ -1819,7 +1817,7 @@ status:
 ~~~ http
 GET /ssf/status?stream_id=f67e39a0a4d34d56b3aa1bc4cff0069f HTTP/1.1
 Host: transmitter.example.com
-Authorization: Bearer zzzz
+Authorization: Bearer eyJ0b2tlbiI6ImV4YW1wbGUifQo=
```
tulshi, line 1820: putting an actual token in GitHub sometimes causes GitHub to give alerts about stored tokens. Are you sure this value won't cause that alert?

TakahikoKawasaki: @tulshi All other examples in the SSF spec use the value `eyJ0b2tlbiI6ImV4YW1wbGUifQo=` as an example access token. If you are concerned about it, all other example access tokens in the SSF spec need to be replaced with `zzzz`.

tulshi (approved):

FragLegs (approved):

PR #286, opened 2025-07-31
Prepare specs for CAEP draft 5, RISC draft 3, and SSF draft 5

FragLegs:
We want this round of implementer's draft to become the final v1. We have added a few non-normative changes during the review period for draft 4 (typo fixes and examples) that we would like to have included in that v1. Mike Jones has suggested that we publish a draft 5 and make it part of the current vote instead of trying to restart the process. There is precedent for this action.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@
 .refcache
 *.xpr
 scratch.md
+openid-caep-1_0.xml
+openid-sharedsignals-framework-1_0.xml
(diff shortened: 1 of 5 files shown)


tulshi (approved):

appsdesh (commented):
Should these version chnages be done after cutting a release branch? Decrementing versions look odd and defeating the previous purpose which incremented it?!?

iamseanodentity (approved):

PR #314, opened 2025-12-16
removed status endpoint requirement

tulshi:
Issue #318


appsdesh (approved):

tulshi:
The status endpoint is required by few organizations that depend on SSF / CAEP

appsdesh:
@atultulshi CAEP interop profile does not suggest stream updates as mandatory stream control operations https://openid.github.io/sharedsignals/openid-caep-interoperability-profile-1_0.html#name-stream-control 

That would render the stream status unnecessary for establishing interoperability. I am reopening this to discuss on the Tuesday call

thomasdarimont:
If we remove the status endpoint requirement here, then I need to skip the tests for the status endpoint in the conformance tests if the caep interop profile is selected.

I think reading the stream status should still be part of the CAEP interop profile, to let a receiver know if they can expect new events form a stream or not. Therefore I suggest to keep the status endpoint section.

(I created an issue for this #318)

Issue #131, opened 2023-12-06
CAEP device management change is not captured in the existing events


appsdesh:
In the world of device management, the device management signal is very important. It does not directly map to the device compliance status. 

Device compliance status could be the result of evaluating the compliance policy (output of a policy decision engine). In certain cases, device management status maybe needed to evaluate policies (input to the policy decision engine).

Given that a managed device may or may not be compliant and this information is important in the continuous access evaluation. 

I am proposing following change to the device compliance CAEP event 

### Option 1 - Management status via new keys

1. Mark existing `current_status` and `previous_status` as optional to avoid mandating compliance status
2. Add additional 2 keys `current_management_status` and `previous_management_status` as OPTIONAL fields to communicate current and previous management status.  

Pros - 
1. Transmitter is able to combine compliance and management changes in the same event, helpful in the scale. 

Cons -
1. Increased event size in the case when both compliance and management status is conveyed

### Option 2 - Extend the values of existing keys

1. Allow new values in the `current_status` and `previous_status` keys 
** managed
** not-managed

Pros -
1. No size increase for the event, reusing the keys
3. Transmitter is able to combine compliance and management changes in the same event, helpful in the scale.

Cons -
1. May not be able to represent management change and compliance change together, due to key reuse

### Option 3 - A new event called device management change

Define a new event as device management change to solely communicate device management change. This new event will have `current_status` and `previous_status` keys and values defined in the **Option 2**

Pros -
1. Cleaner separation of two constructs 

Cons -
1. A new event to fire and consume for Tx/Rx pair

timcappalli:
I'd prefer option 3. Device management status and device compliance are very different signals.

Device management status is often one piece of context that is used to determine device compliance.

appsdesh:
I agree with @timcappalli. 

My preference is in the following order  -
Option 3 (Most favorable)
Option 1
Option 2 (least)

tulshi:
I agree that option 3 is a good one, although I'd like to call the event "device management status" (IMO we should stop using "change" in event names, because all events indicate some change). I feel the two events "device compliance" and "device management" are different, so having both of them co-exist is fine.

tulshi:
@appsdesh could you please verify if this issue is still current?

ysarig75:
Reopening the issue per the discussion we had in the 04/14/26 meeting: https://hackmd.io/@oidf-wg-sse/wg-meeting-20260414

martingalloar:
[...]

PR #329, opened 2026-04-21
Add an option to report device management status change to device compliance change signal

ysarig75:
Add an 'unmanaged' option to current_status in device compliance change event.

### DIFF
--- a/openid-caep-1_0.md
+++ b/openid-caep-1_0.md
@@ -737,7 +737,7 @@ be one of the following strings:
 >
 > * `compliant`
 > * `not-compliant`
-> * `unmanaged`: The device is no loner being managed and additional compliance change events may not be sent for this device. The device compliance state is unknown
+> * `unmanaged`: The device is no longer being managed, and additional compliance change events may not be sent for this device. The device’s compliance state is unknown
 
 When `event_timestamp` is included, its value MUST represent the time at which
 the device compliance status changed.
--- a/openid-caep-1_0.md
+++ b/openid-caep-1_0.md
@@ -37,6 +37,13 @@ contributor:
         contribution: |
           Apoorva defined the `risk-level-change` event.
 
+contributor:
+      -
+        ins: Y. Sarig
+        name: Yair Sarig
+        org: Omnissa
+        email: ysarig@omnissa.com
+
 normative:
   ISO-IEC-29115:
     target: https://www.iso.org/iso/iso_catalogue/catalogue_tc/catalogue_detail.htm?csnumber=45138
@@ -730,6 +737,7 @@ be one of the following strings:
 >
 > * `compliant`
 > * `not-compliant`
+> * `unmanaged`: The device is no loner being managed and additional compliance change events may not be sent for this device. The device compliance state is unknown
 
 When `event_timestamp` is included, its value MUST represent the time at which
 the device compliance status changed.
--- a/openid-caep-1_0.md
+++ b/openid-caep-1_0.md
@@ -729,6 +729,7 @@ triggered the event. This MUST be one of the following strings:
 >
 > * `compliant`
 > * `not-compliant`
+> * `unknown`
 
 current_status
 
--- a/openid-caep-1_0.md
+++ b/openid-caep-1_0.md
@@ -36,8 +36,6 @@ contributor:
         email: apoorva.deshpande@okta.com
         contribution: |
           Apoorva defined the `risk-level-change` event.
-
-contributor:
       -
         ins: Y. Sarig
         name: Yair Sarig
--- a/openid-caep-1_0.md
+++ b/openid-caep-1_0.md
@@ -1,9 +1,8 @@
 ---
-title: OpenID Continuous Access Evaluation Profile 1.0
+title: OpenID Continuous Access Evaluation Profile 1.1
 
 abbrev: CAEP-Spec
 docname: openid-caep-1_0
-date: 2025-08-29
 
 ipr: none
 cat: std
@@ -712,7 +711,7 @@ Event Type URI:
 
 `https://schemas.openid.net/secevent/caep/event-type/device-compliance-change`
 
-Device Compliance Change signals that a device's compliance status has changed.
+Device Compliance Change signals that a device's compliance or managed status has changed.
 
 The actual reason why the status change occurred might be specified with the
 nested `reason_admin` and/or `reason_user` claims made in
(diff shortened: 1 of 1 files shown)


openid-caep-1_0.md
```diff
@@ -730,6 +738,7 @@ be one of the following strings:
 >
 > * `compliant`
 > * `not-compliant`
+> * `unmanaged`: The device is no longer being managed, and additional compliance change events may not be sent for this device. The device’s compliance state is unknown
```
martingalloar: I believe using "unmanaged" here might be confusing, and also from a flow/transitioning perspective can potentially lead to dead end scenarios if the previous status is "unknown".

I'd suggest normalizing into the same value (either "unknown" or "unmanaged") for both previous and current status.

ysarig75: @martingalloar `previous_status` and `current_status` are not symmetrical. I added `unknown` to `previous_status` because it is possible that the transmitter doesn't have information on the previous status to send. For `current_status`, `unknown` is not very useful to send as it is not clear what a receiver will do with such a status.

thomasdarimont: Since the `previous_status` is `required` how about using "unmanaged" as the initial "null" value.
With this we might not need the "unknown" value anymore IMHO.

So a device could transition from (initially) unmanaged -> not-compliant -> compliant -> unmanaged (again once removed / retired).

ysarig75: @thomasdarimont The problem is not the transition of the device status (which is communicate using `current_status` field but the fact that the previous state is currently required and the transmitter doesn't necessarily know what was the `previous_state`. So a transmitter may send a `current_status=compliant` and a a month later, the device become unmanaged and the transmitter want to send a `current_status=unmanaged` signal but at that point, the transmitter may not know what was the `previous_state`.

openid-caep-1_0.md
```diff
@@ -37,6 +37,13 @@ contributor:
         contribution: |
           Apoorva defined the `risk-level-change` event.
 
+contributor:
```
appsdesh: Incorrectly formatted

openid-caep-1_0.md
appsdesh: @ysarig75, what if the device becomes managed from unmanaged? How would it be represented, as `previous_status` does not have `unmanaged`

My alternate proposal is to add a new key `current_management_status` to the event with the values `unknown`, `managed`, `unmanaged`

This way, we do not need to worry about Tx/Rx that do not know how to process new values in the existing claims.

openid-caep-1_0.md
ysarig75: @appsdesh Can you explain what is the formatting issue?

openid-caep-1_0.md
ysarig75: @appsdesh I like adding `current_management_status` as you described but we will also need to add `unknown` to both `previous_status` and `current_status` because they are mandatory (you can't just send a  `current_management_status`  in a signal).

openid-caep-1_0.md
appsdesh: contributor section already exists

ysarig75: Fixed

appsdesh:
@ysarig75 related issue and proposal I had long back here https://github.com/openid/sharedsignals/issues/131

jmarchesini (approved):

openid-caep-1_0.md
tulshi: Please change title to "1.1 - Draft"

tulshi: How do we ensure the date is automatically updated? @jischr you had investigated this, right?

jischr: Yep - remove line 6 and then the date gets automatically populated when the html and xml are built

openid-caep-1_1.md
```diff
@@ -707,30 +711,40 @@ Event Type URI:
...
+> OPTIONAL, JSON string: the compliance status prior to the change that
 triggered the event. This MUST be one of the following strings:
 >
 > * `compliant`
 > * `not-compliant`
 
 current_status
 
-> REQUIRED, JSON string: the current status that triggered the event. This MUST
+> OPTIONAL, JSON string: the current status that triggered the event. This MUST
```
tulshi, line 740: Should we say that this value is optional if the management status is "unmanaged", but required otherwise? Or we could add "unknown" as a value to this enum (and the previous status one), so that the fields are required, but the value can be set to unknown if the device is not managed.

appsdesh: We should not change the current status to optional, as doing that would break backward compatibility, as receiver logic most likely uses the current status (and not necessarily the previous status)

openid-caep-1_1.md
```diff
@@ -707,30 +711,40 @@ Event Type URI:
...
-Device Compliance Change signals that a device's compliance status has changed.
+Device Compliance Change signals that a device's compliance or managed status has changed.
 
 The actual reason why the status change occurred might be specified with the
 nested `reason_admin` and/or `reason_user` claims made in
 {{optional-event-claims}}.
 
 ### Event-Specific Claims {#device-compliance-change-claims}
 
+The transmitter MUST send at least one of the current_status or current_management_status claims.
```
appsdesh, line 728: Preferably, we should not change the existing understanding of compliance status, only make additive changes to the management status

openid-caep-1_1.md
ysarig75: I agree that adding unknown option to current_status and leaving it a required may be a better option.

openid-caep-1_1.md
ysarig75: We decided to have a different event type URI for updated events so this is no longer an issue

openid-caep-1_1.md
ysarig75: We decided to have a different event type URI for updated events so this is no longer an issue

openid-caep-1_0.md
jogu: That's already been handled in https://github.com/openid/sharedsignals/pull/310 - I'd suggest finishing that MR first.

openid-caep-1_1.md
```diff
@@ -705,32 +713,42 @@ the assurance level changed.
 
 Event Type URI:
 
-`https://schemas.openid.net/secevent/caep/event-type/device-compliance-change`
+`https://schemas.openid.net/secevent/caep/event-type/v1.1/device-compliance-change`
```
jogu, line 718: Is there an issue with more background? If so could it be linked from the pull request description please?

For a 1.1 version of the spec (people generally expect specification versions to follow semantic versioning), I would expect that the 1.0 events would be retained, and the 1.1 events added in addition. Otherwise 1.0 receivers might not work with 1.1 transmitters. Removing an event from the profile (which is effectively what this is doing) should in my opinion be reserved for a 2.0 version of the spec.

jogu (changes requested):

openid-caep-1_1.md
ysarig75: I added a comment to clarify that this spec doesn't deprecate 1.0 events and they can continue to be used.

openid-caep-1_1.md
```diff
@@ -243,6 +247,12 @@ The base URI for CAEP event types is:
 
 `https://schemas.openid.net/secevent/caep/event-type/`
 
+Events that were added or updated by this spec will use the following base URI:
+
+`https://schemas.openid.net/secevent/caep/event-type/v1.1`
+
+This spec does not deprecate 1.0 events, you can continue to use those.
```
jogu, line 254: I think we should clarify who can use them, although maybe some of that belongs in the interop profile. But for here it could be useful to e.g. say if we recommend that transmitters should continue to support the 1.0 events so that 1.0 receivers interoperate with them.

openid-caep-1_1.md
jogu: Thanks, that helps!

I think there's 3 options:

1. The spec includes all events that are valid to use in 1.1, including those defined in 1.0
2. The spec only includes only the latest version of each event, which might be the 1.0 version of the 1.1 version (the current position in this PR I believe)
3. The spec only includes events that are added/updated in 1.1

I think any of the 3 would be okay, but it should be a clear policy.

PR #334, opened 2026-05-26
Closes #332 - Fix scopes section

### LINKED ISSUE #332: scope section wording is odd
https://openid.github.io/sharedsignals/openid-caep-interoperability-profile-1_0.html#section-2.7.2 says:

> The following scopes MUST be supported:
>
> * An OAuth [[RFC6749](https://openid.github.io/sharedsignals/openid-caep-interoperability-profile-1_0.html#RFC6749)] Authorization Server that is used to issue tokens to SSF Receivers, MUST reserve the scopes for the SSF endpoints with the prefix of ssf

That first bullet doesn't actually contain a scope that must be supported, so the 'The following scopes MUST be supported:' is not correct - probably that 1st line could just be removed.

I'm not sure exactly what the intent of the first bullet actually is, I think it's basically saying "the AS mustn't use any scopes beginning with ssf for non-use things". If that's the intent it should probably say `ssf.`, as I think that's the actual prefix used and including the '.' potentially avoids clashes with existing scopes.

jischr:
Addressing Issue #332 by fixing the wording.


tulshi (approved):

appsdesh (changes requested):
The new wording makes the scores mutually exclusive. The managed scope should also allow all CRUD operations.

jischr:
@appsdesh thanks for the comment. can you take a look at the updated wording and let me know what you think?

appsdesh (approved):

appsdesh:
> @appsdesh thanks for the comment. can you take a look at the updated wording and let me know what you think?

LGTM! thanks

iamseanodentity (approved):
ship it

PR #335, opened 2026-06-02
Update scope wording and Fix #327

### LINKED ISSUE #327: Clarify how a Receiver discovers the Authorization Server from a Transmitter URL
In CAEPIOP 1.0 2.7.2 ("OAuth Scopes") explicitly mentions OAuth Protected Resource Metadata (OPRM, RFC 9728) for 
**scope discovery**, but neither CAEPIOP 2.7 nor SSF 1.0 7 explicitly says how a Receiver should  discover the **Authorization Server** itself, 
given only a Transmitter URL. 

In practice OPRM's `authorization_servers` field is the only in-band mechanism that does this, but the spec doesn't say so directly.

I think it would help interop if CAEPIOP 2.7 stated the expected discovery path explicitly.

Currently in the conformance tests, we require the configuration of a token endpoint if "Authentication Variant" is set to "dynamic".
If we could optionally leverage the oauth resource server from the OPRM, we could discover the token endpoint through the OAuth Authorization Server metatdata if present.

jischr:
This PR address two things:
1. It adds more clarification to PR #334 by outlining usage of `ssf.` prefix
2. Fixes issue #327 by adding "The method in which the Receiver determines which Authorization
Server to use is out of scope."

### DIFF
--- a/openid-caep-interoperability-profile-1_0.md
+++ b/openid-caep-interoperability-profile-1_0.md
@@ -367,7 +367,8 @@ different entity than the SSF Transmitter.
 
 * The Authorization Server MAY distribute discovery metadata (such as the
 authorization endpoint) via Authorization Server Metadata as specified in
-{{RFC8414}}
+{{RFC8414}}. The method in which the Receiver determines which Authorization
+Server to use is out of scope.
 * The Authorization Server MUST support at least one of the following to issue a
 short-lived access token to the Receiver:
   * client credentials grant {{RFC6749}} Section 4.4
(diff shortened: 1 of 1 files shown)


jogu (approved):

openid-caep-interoperability-profile-1_0.md
```diff
@@ -393,7 +394,9 @@ Server MUST return errors as per Section 3.1 of {{RFC6750}}
 
 ### OAuth Scopes
 
-SSF Receivers MUST reserve the scopes with a prefix of `ssf.` for SSF endpoints.
+Scopes values beginning with ‘ssf.’ are reserved and MUST only be defined by
```
appsdesh, line 397: > MUST only be defined by the SSF specifications.

Is there any enforcement mechanism, if not, I will opt for removing this language

jogu: This language has been there for a long time btw, this PR doesn't newly add it.

The intention (as I understand the text that was originally put in the spec by someone)  is to say 'scopes beginning ssf. are reserved and you mustn't create a `ssf.myweirdthing` scope for your own extension as that might conflict with a future extension of the core SSF specs.

jogu: It could be tested in the certification tests by scanning scopes_supported in the AS metadata for values beginning `ssf.` and making sure they're all values we recognise.

appsdesh (approved):

Issue #330, opened 2026-05-07
Add `UNKNOWN` as a valid value for `current_level` / `previous_level` in `risk-level-change`


jtanios:
### Context

The [risk-level-change event](https://openid.github.io/sharedsignals/openid-caep-specification-1_0.html#name-risk-level-change), introduced in #200, defines `current_level` as REQUIRED with values restricted to `LOW`, `MEDIUM`, or `HIGH`. There is no way for a Transmitter to communicate "I do not currently have enough signal to assess this subject's risk."

This conflates two distinct states:

- **HIGH:** the Transmitter has high confidence the subject is risky (e.g., suspected non-authorized user, compromised credential, malicious software).
- **No assessment available:** the Transmitter previously had a risk level but signal has gone stale, the subject is idle, or upstream data sources are unavailable.

Today, Transmitters facing the second case must choose between three bad options:

1. Emit `HIGH` (fail-closed). This is semantically incorrect, it overstates the Transmitter's confidence and can drive unwarranted remediations like forced re-auth or session revocation.
2. Continue reporting the last known level. The Receiver has no way to know the assessment is stale.
3. Stop emitting events. The Receiver cannot distinguish "signal lost" from "no change."

### Proposal

Add `UNKNOWN` as a permitted value for `current_level` and `previous_level`.

- `current_level: UNKNOWN` indicates the Transmitter does not currently have sufficient signal to assess the subject's risk.
- `previous_level: UNKNOWN` is semantically equivalent to omission (already defined as "unknown to the Transmitter") but allows explicit transitions, e.g., `UNKNOWN -> LOW` when signal is re-established.

### Example Scenarios

- **Continuous behavioral authentication:** A Transmitter assessing a user via behavioral biometrics loses signal when the user goes idle or the agent disconnects. Transitioning to `UNKNOWN` lets Receivers apply their own policy (e.g., step-up on next sensitive action) without the Transmitter falsely asserting risk.
- **Device posture:** A device's compliance agent stops checking in. The Transmitter can emit `UNKNOWN` rather than guess, letting Receivers decide whether absence of signal warrants remediation.
- **Upstream provider outage:** A Transmitter ingesting risk from a third-party provider can communicate signal loss without retracting prior assessments or fabricating a level.

### Event-Specific Claims (updated)

1. `current_level` - REQUIRED. Value MUST be one of `LOW`, `MEDIUM`, `HIGH`, `UNKNOWN`.
[...]

PR #353, opened 2026-09-01
Editorial fixes to the Risk Level Change use case

mike-kiser-sp:
Follow-ups to #349, all confined to the section it added. No normative change.

### 1. Defer `current_level` values to CAEP instead of enumerating them

The merged text hardcodes `(LOW, MEDIUM, HIGH)`. #330 proposes adding `UNKNOWN` to that enum in CAEP, which would leave the profile carrying a stale list.

This is also inconsistent *within the section as merged*: `principal` and `previous_level` both say "all allowable values" and defer to CAEP, as do Sections 3.1-3.3. Only `current_level` enumerates.

### 2. `Section 3.8.1` rather than the section sign

The document sets `coding: us-ascii`, so `§` is transliterated by kramdown. The published page currently reads **"1.0 S.3.8.1"** — see [Section 3.4](https://openid.github.io/sharedsignals/openid-caep-interoperability-profile-1_0.html#section-3.4). The `{{XXX}} Section N` form is used 18 times elsewhere in the document.

Both occurrences are in the newly added section.

### 3. Trailing whitespace on three definition-list terms

`current_level `, `previous_level `, `reason_admin ` — kramdown treats trailing whitespace as a line break.

### 4. Trailing newline at end of file

The file ended mid-line, so any future edit to the document history would carry a spurious extra line in its diff.

---

Verified with `kramdown-rfc2629`: builds clean, and the `{{CAEP}}` xref resolves to `<xref target="CAEP"/> 1.0 Section 3.8.1` in both places.

<sub>Replaces #352, which could not be reopened after a force-push.</sub>


thomasdarimont:
The change looks good to me.

